PR #18, opened 2022-05-29
On block placement correct orientation based of ships orientation

ewoudje:
As noted in the title, this should make furnaces etc rotate towards the player irrelevant of the ships rotation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,11 @@ buildscript {
     configurations.classpath {
         resolutionStrategy.activateDependencyLocking()
     }
+
+    repositories {
+        mavenCentral()
+        mavenLocal()
+    }
 }
 
 plugins {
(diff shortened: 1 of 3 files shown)


deltanedas:
no idea who you are but this looks epic :+1:

StewStrong:
I think the current code is supposed to teleport (and rotate) the player to the ship blocks before placing the block, though I need to check this.

If that's the case, then we should probably fix this issue in the existing code that teleports/rotates the player.

ewoudje:
> I think the current code is supposed to teleport (and rotate) the player to the ship blocks before placing the block, though I need to check this.
> 
> If that's the case, then we should probably fix this issue in the existing code that teleports/rotates the player.

Oh is there existing code that does this? I was trying to get used to the codebase by implementing this.

And btw me just random dude that liked the mod and saw there was development for a 2 that was recently updated :P

ewoudje:
So as discussed on discord this intercepts at every block interaction and transforms the player to the ship space for the block building

StewStrong:
Thank you so much! Very cool :)

PR #56, opened 2022-06-08
ShipableBlockEntity Interface

ewoudje:
So im making this PR so that we can discuss api's, and i will make a PR for every api
This will also be clearer then discord since in #general everything just get blasted over by other messages

So about this PR i think what it should do is quite clear based on the commit.
The reason for this API is to be used together with another API idea i had: storing data on a per ship basis.


deltanedas:
ok but who is the blockentity being shipped with :flushed:

ewoudje:
Since there are new api ideas im making a new one and closing this one

PR #59, opened 2022-06-09
Ship API

ewoudje:
Main PR: https://github.com/ValkyrienSkies/vs-core/pull/11

This includes a minecraft extension of ShipProvider specific for block entities
And also a kotlin util for ship data management


StewStrong:
Looks pretty epic! Though, waiting on Ruby's thoughts before merging.

common/src/main/kotlin/org/valkyrienskies/mod/api/ShipFields.kt
```diff
@@ -0,0 +1,31 @@
+package org.valkyrienskies.mod.api
```
Rubydesic, line 1: This entire file should be in vs-core

common/src/main/kotlin/org/valkyrienskies/mod/api/BlockEntityShipProvider.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * Gets called on block entity creation and relocation
+     *  the actual new [BlockPos], [Ship] and [Level] shall be assigned
+     *  on this [BlockEntity] after this call
+     * @param newPos
+     * @param newLevel
+     * @param newShip
+     */
+    fun shipChange(newPos: BlockPos, newLevel: Level, newShip: ShipObject) {}
```
Rubydesic, line 21: maybe name it 'onShipChange' to indicate it's meant to be called by VS and not by the addon.

Rubydesic (changes requested):

Rubydesic (approved):

Issue #32, opened 2022-06-05
Send delete voxel terrain updates to Krunch when a chunk is unloaded


StewStrong:
Currently we don't do this, which is effectively a memory leak because Krunch won't forget about unloaded chunks. Need to fix this eventually, definitely before Eureka.

StewStrong:
See https://github.com/ValkyrienSkies/vs-core/pull/14 and https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/72

StewStrong:
Done in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/074cd299782975c9dbbe89ac8c144d8db5f5ecfe

PR #72, opened 2022-06-26
Deleted unloaded blocks from Krunch

StewStrong:
Working on this issue https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/32

Currently there is a bug which causes Krunch to switch from 1 sphere to 8 spheres per block when a certain number of blocks is loaded. This fix makes this bug more apparent as the switch is now constantly occurring as chunks are loaded/unloaded.

This switch causes ships to bounce up and down, which is very silly.

TODO: Currently this only unloads terrain from the ground body; need to make it unload chunks from ships that are deleted/unloaded as well.


StewStrong:
We should probably finish this after the networking branch is merged.

StewStrong:
I remade this change off of mainline because its easier https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/074cd299782975c9dbbe89ac8c144d8db5f5ecfe

PR #79, opened 2022-06-29
 feature/networking - rewrite networking and ship loading

Rubydesic:
see also https://github.com/ValkyrienSkies/vs-core/pull/16


common/src/main/java/org/valkyrienskies/mod/mixin/client/world/MixinClientChunkCache.java
```diff
@@ -71,6 +74,8 @@ private void preLoadChunkFromPacket(final int x, final int z, final ChunkBiomeCo
     @Inject(method = "drop", at = @At("HEAD"), cancellable = true)
     public void preUnload(final int chunkX, final int chunkZ, final CallbackInfo ci) {
         shipChunks.remove(ChunkPos.asLong(chunkX, chunkZ));
+        ((IVSViewAreaMethods) ((LevelRendererAccessor) ((ClientLevelAccessor) level).getLevelRenderer()).getViewArea())
```
StewStrong, line 77: Nice casting 😂

common/src/main/java/org/valkyrienskies/mod/mixinducks/server/IPlayerProvider.java
```diff
@@ -0,0 +1,8 @@
+package org.valkyrienskies.mod.mixinducks.server;
+
+import java.util.UUID;
+import org.valkyrienskies.core.game.IPlayer;
+
+public interface IPlayerProvider {
+    IPlayer getPlayer(UUID uuid);
```
StewStrong, line 7: Should this return a nullable value?

PR #80, opened 2022-06-30
feature/dynamic-lights: Lights on the ship affect the world and vice versa

Rubydesic:
Still a little laggy and only works from ship->world at the moment.

Needs ship aabbs and other performance optimizations


ewoudje:
Out of date, from a prev era

PR #84, opened 2022-07-02
An util for assembling, so that it also move block entities

ewoudje:
Maybe in the future also add a floodfill, so that addons can use it to assemble ships.
Not gonna do that cus my floodfill is crap

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/util/RelocationUtil.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/util/RelocationUtil.kt
@@ -1,0 +1,26 @@
+package org.valkyrienskies.mod.common.util
+
+import net.minecraft.core.BlockPos
+import net.minecraft.world.level.Level
+import net.minecraft.world.level.chunk.LevelChunk
+
+// the from and the to can be local or global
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/util/RelocationUtil.kt
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.world.level.Level
+import net.minecraft.world.level.chunk.LevelChunk
+
+// the from and the to can be local or global
+fun relocateBlock(fromChunk: LevelChunk, from: BlockPos, toChunk: LevelChunk, to: BlockPos) {
+    val state = fromChunk.getBlockState(from)
+    val entity = fromChunk.getBlockEntity(from)
+
+    toChunk.setBlockState(to, state, false) // TODO should isMoving be false?
+    entity?.let {
```
StewStrong, line 13: I think the safest way to do this is to save the entity to a CompoundTag, then manually edit the position thats saved, and then load the new BlockEntity from the modified tag.

We can also add a relocation API to allow mods to fix bugs with relocated blocks.

ewoudje:
This PR also fixes the blockentity distance checks, and maybe others.. (see MixinEntity)
And everything works now is ready to be merged

StewStrong (approved):
Freaking sweat!

PR #98, opened 2022-07-17
Feature/entity ship mounting

StewStrong:
TODO:
* Need to add an api for addons to use this
* Need to adjust the 3rd person camera distance
Also see https://github.com/ValkyrienSkies/vs-core/pull/25


common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    override fun getControllingPassenger(): Entity? {
+        return this.passengers.getOrNull(0)
+    }
+
+    override fun getAddEntityPacket(): Packet<*> {
+        return ClientboundAddEntityPacket(this)
+    }
+
+    override fun addPassenger(passenger: Entity) {
```
ewoudje, line 69: This can be removed *normaly*

fabric/src/main/kotlin/org/valkyrienskies/mod/fabric/common/ValkyrienSkiesModFabric.kt
```diff
@@ -49,4 +73,15 @@ class ValkyrienSkiesModFabric : ModInitializer {
                 }
             })
     }
+
+    private fun registerBlockAndItem(registryName: String, block: Block) {
```
ewoudje, line 77: Shall i write a registration util, for addons and vs itself?

ewoudje (approved):
Spawning the entity and deleting the moment it isn't needed anymore is way better then the method i use 👍 

About how should have api for this.. We could add a attachment when somebody is sitting on a seat called SeatedPlayerController or smthng and that is used then by other attachments as thing for input and checking if there is even somebody sitting on the ship rightnow

Oh and then a boolean to control if its a passenger seat, so not adding the attachment.

fabric/src/main/kotlin/org/valkyrienskies/mod/fabric/common/ValkyrienSkiesModFabric.kt
StewStrong: Maybe? I feel like the Arch API does this better, but I didn't use it since we don't want VS2 to depend on the Arch API.

ewoudje: Ye arch api can do that too but its wriiten for java, we can make cool kotlin registering 😏 that together with not depending on arch api. honestly depending on arch loom already hurts enough (altough i have yet to have problem with api but)

PR #138, opened 2022-08-08
Refactor/more while were at it

ewoudje:
https://github.com/ValkyrienSkies/vs-core/pull/32


ewoudje:
https://github.com/ValkyrienSkies/vs-core/issues/33

PR #140, opened 2022-09-09
Feature/entities

ewoudje:
Entities get handled, I made an interface called VSEntityHandler it has a couple of methods that are neededd to have a
ShipyardEntityHandler (aka it lives in shipyard renders in world) or WorldEntityHandler. 
There should be more added later but for now this is good enough.

So wich handler should we use for wich entity? well since we need mod compat ive made a datapack whereof u can configure this and thus easy for other mods to support or modpacks.

The default is WorldEntityHandler wich teleports all entities out the ship and in worldspace.


StewStrong:
Unverified commit do be sussin

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/entity/MixinEntityRenderDispatcher.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Mixin(EntityRenderDispatcher.class)
+public class MixinEntityRenderDispatcher {
+
+    @Inject(method = "render",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/client/renderer/entity/EntityRenderer;render(Lnet/minecraft/world/entity/Entity;FFLcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/renderer/MultiBufferSource;I)V",
+            shift = At.Shift.BEFORE),
+        locals = LocalCapture.CAPTURE_FAILHARD)
```
StewStrong, line 27: Capture those locals!

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/entity/MixinEntityRenderDispatcher.java
```diff
@@ -0,0 +1,66 @@
...
+        cancellable = true
+    )
+    void shouldRender(final Entity entity, final Frustum frustum,
+        final double camX, final double camY, final double camZ,
+        final CallbackInfoReturnable<Boolean> cir) {
+        if (!cir.getReturnValue()) {
+            final ClientShip ship =
+                (ClientShip) VSGameUtilsKt.getShipObjectManagingPos(entity.level, entity.blockPosition());
+            if (ship != null) {
+                // TODO actually check
```
StewStrong, line 60: Can you actually check pls :)

ewoudje: i forgor 💀

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/entity/MixinEntityRenderDispatcher.java
```diff
@@ -0,0 +1,88 @@
...
+        cancellable = true
+    )
+    void shouldRender(final Entity entity, final Frustum frustum,
+        final double camX, final double camY, final double camZ,
+        final CallbackInfoReturnable<Boolean> cir) {
+        if (!cir.getReturnValue()) {
+            final ClientShip ship =
+                (ClientShip) VSGameUtilsKt.getShipObjectManagingPos(entity.level, entity.blockPosition());
+            if (ship != null) {
+                AABB aABB = entity.getBoundingBoxForCulling().inflate(0.5);
```
StewStrong, line 71: I think theres already a function to transform AABB through a ShipTransform

common/src/main/java/org/valkyrienskies/mod/mixin/entity/MixinEntity.java
```diff
@@ -295,6 +296,17 @@ private void preDistanceToSqr(final double x, final double y, final double z,
         cir.setReturnValue(VSGameUtilsKt.squaredDistanceToInclShips(Entity.class.cast(this), x, y, z));
     }
 
+    /**
+     * @author ewoudje
+     * @reason use vs2 handler to handle this method
+     */
+    @Overwrite
```
StewStrong, line 303: We can't use @Overwrite because it breaks Eureka dev environment because of a bug in Architectury. So we'll have to replace this with an @Inject.

common/src/main/kotlin/org/valkyrienskies/mod/common/PlayerUtil.kt
```diff
@@ -45,9 +45,9 @@ object PlayerUtil {
             val position = shipMatrix.transformPosition(
                 player.position().toJOML()
             )
-            val yaw = atan2(direction.x, -direction.z) // yaw in radians
-            val pitch = asin(-direction.y)
-            player.yRot = (yaw * (180 / Math.PI)).toFloat() + 180
+            val yaw = -atan2(direction.x, direction.z)
+            val pitch = -atan2(direction.y, sqrt((direction.x * direction.x) + (direction.z * direction.z)))
```
StewStrong, line 49: Pitch is always between -90 and 90, so I don't think atan2 is necessarily what we want here, because it can return numbers outside of that range.

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/entity/MixinEntityRenderDispatcher.java
Rubydesic: AABBdc#transform

ewoudje: Then i need to convert it from mc to joml and back :/, but ig its cleaner this way

ewoudje:
It was doing wierd until i used this so :person_shrugging:

common/src/main/kotlin/org/valkyrienskies/mod/common/PlayerUtil.kt
ewoudje: It was doing wierd until i used this so :person_shrugging:

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/entity/MixinEntityRenderDispatcher.java
ewoudje: And would fix scaling

StewStrong (approved):
Is gud!

PR #157, opened 2022-09-29
Add fire spread between ship and world

copperwarrior:
Fire and lava will spread fire between the ship and the world.

Solution for lava is fairly clean, but the tick for the fire block does too much to enable a clean solution.


common/src/main/java/org/valkyrienskies/mod/mixin/world/level/block/FireMixin.java
```diff
@@ -0,0 +1,112 @@
...
+@Mixin(FireBlock.class)
+public class FireMixin {
+    @Unique
+    private boolean isModifyingFireTick = false;
+
+    @Shadow
+    @Final
+    public static IntegerProperty AGE;
+
+    //TODO make mixin less intervening
```
StewStrong, line 30: Delete todo pls

PR #162, opened 2022-10-16
Replace BlockGetter#clip

ewoudje:
So copper asked me if BlockGetter#clip was replaced, and it wasn't.
This fixes allot of inherent issues, such as projectiles. 
We probably can remove also other fixes in place of this.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -27,9 +27,6 @@
       <option name="IMPORT_NESTED_CLASSES" value="true" />
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
-    <ScalaCodeStyleSettings>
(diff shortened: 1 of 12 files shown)


copperwarrior:
Confirmed fixes the following issues:
 - Mobs seeing entities and targets through ship walls
 - Projectiles not colliding with ships, arrows still janky but projectile branch now obsolete
 - F5 not colliding with ship walls, should make exception for this in custom camera code
 - Entity damage from explosions not being blocked by ship walls
Does not fix:
 - Falling sand not functioning properly
 - Explosions without explosion mixin
Breaks:
 - Nothing that I could find

Additionally, things that worked before like the pick function for entities and items placeable on ships now work better.

ewoudje:
Only arrows are a bit jitterish but that can be addressed by a separate branch
Explosions are kinda fixed:tm:
It now only raytraces once, no changes to getEntities for now.
The crash causing clip now throws a warning and returns a miss. (Just like how getEntities throw errors for too big AABB's)

Rubydesic (approved):

PR #167, opened 2022-10-24
Fix create

ewoudje:


Rubydesic:
standing issues: 

- everything disappears if you reload the world (maybe tile entities are getting created before the ship object is created, so #createRaw isn't getting redirected?
- remove MixinTileInstanceManagerDuck which does nothing
- remove dependency on create
- Make flywheel dependency modCompileOnly or modProvided or whatever its called
-

ewoudje:
A extra issue: When hovering with the wrench it should display information on certain blocks it doesn't do this on ships.

Rubydesic:
Clear to merge after #169

PR #168, opened 2022-10-24
Fixed getPosStandingOnFromShips and moved entity collision mixins

ewoudje:
Moved entity collision to a feature and fixed the incorrect posStanding


Rubydesic:
(changed round -> floor)

PR #169, opened 2022-10-24
Make ship assembly seamless; prevent falling through ships

Rubydesic:
Prevents falling through ships and flickering (for nearby ship chunks being assembled - this is dependent on rebuildChunkSync being called - async chunk rebuilds still cause flickering - see feature/better-assembly for a questionable attempt at fixing this)

This is accomplished by sending a **stop chunk updates** packet to the client, which causes the client to queue all incoming packets for that chunk until a **restart chunk updates** packet is received, at which point all the packets are processed at once on the next tick.

This also queues packets for ships that don't exist yet, fixing invisible tile entities after reloading the world in the fix-create branch (#167)


common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/SeamlessChunksManager.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private fun onRestartUpdates(packet: PacketRestartChunkUpdates) {
+        val (chunks) = packet
+
+        chunks.forEach { pos ->
+            stalledChunks.remove(pos.toLong())
+            val packets = queuedUpdates.remove(pos)
+            if (!packets.isNullOrEmpty()) {
+                logger.info("Executing ${packets.size} deferred updates at <${pos.x}, ${pos.z}>")
```
ewoudje, line 74: When merged shouldn't this be a debug to reduce the console spam

fabric/src/main/kotlin/org/valkyrienskies/mod/fabric/common/ValkyrienSkiesModFabric.kt
```diff
@@ -68,6 +70,10 @@ class ValkyrienSkiesModFabric : ModInitializer {
         ValkyrienSkiesMod.init(vsCore)
 
         registerBlockAndItem("test_chair", ValkyrienSkiesMod.TEST_CHAIR)
+        Registry.register(
+            Registry.ITEM, ResourceLocation(ValkyrienSkiesMod.MOD_ID, "ship_assembler"),
+            ValkyrienSkiesMod.SHIP_ASSEMBLER_ITEM
+        )
```
ewoudje, line 76: Shouldn't we have a forge counterpart of this?

common/src/main/java/org/valkyrienskies/mod/mixinducks/feature/seamless_copy/ChunkCompileTaskDuck.kt
```diff
@@ -0,0 +1,8 @@
+package org.valkyrienskies.mod.mixinducks.feature.seamless_copy
+
+import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher.RenderChunk
+
+interface ChunkCompileTaskDuck {
```
ewoudje, line 5: Kotlin in the java source?! Even if it does i think its a better idea to make it java for consistency

PR #177, opened 2022-10-31
Explosion push force for ships

copperwarrior:
Branch includes a gametick force applier and explosion force for ships.
Explosion force is relative to distance to explosion and radius, interpreted as the explosive power.
Explosion force is configurable, relative to a single TNT explosion at the center.


common/src/main/java/org/valkyrienskies/mod/mixin/feature/explosions/MixinExplosion.java
```diff
@@ -72,6 +127,14 @@ private void afterExplode(final CallbackInfo ci) {
         isModifyingExplosion = false;
     }
 
+    @Inject(at = @At("TAIL"), method = "finalizeExplosion")
+    private void afterFinalizeExplosion(final boolean spawnParticles, final CallbackInfo ci) {
+        if (this.level.isClientSide) {
+            return;
+        }
```
Rubydesic, line 134: This mixin doesn't do anything..

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/explosions/ClipContextMixin.java
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraft.world.level.ClipContext;
+import net.minecraft.world.phys.shapes.CollisionContext;
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(ClipContext.class)
+public class ClipContextMixin {
```
Rubydesic, line 13: Can we just edit your code to not pass in null values instead of this mixin?

common/src/main/java/org/valkyrienskies/mod/mixin/feature/explosions/MixinExplosion.java
```diff
@@ -49,6 +60,50 @@ public abstract class MixinExplosion {
     @Inject(at = @At("TAIL"), method = "explode")
     private void afterExplode(final CallbackInfo ci) {
         if (isModifyingExplosion) {
+            // Custom forces
+            final Vector3d originPos = new Vector3d(this.x, this.y, this.z);
```
Rubydesic, line 64: Can we move this pushing code into a separate private method at least (make the private method @Unique)

common/src/main/java/org/valkyrienskies/mod/mixin/feature/explosions/MixinExplosion.java
```diff
@@ -15,7 +22,11 @@
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.valkyrienskies.core.api.ServerShip;
 import org.valkyrienskies.mod.common.VSGameUtilsKt;
+import org.valkyrienskies.mod.common.config.VSGameConfig;
+import org.valkyrienskies.mod.common.util.GameTickForceApplier;
+import org.valkyrienskies.mod.common.util.VectorConversionsMCKt;
 
 @Mixin(Explosion.class)
```
Rubydesic, line 31: I think a decent portion of this could go into vs core, but we can leave it here for now.

Rubydesic (approved):

Rubydesic:
Sorry meant to request changes but I have fat fingers and pressed approve. Otherwise looks good though

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/explosions/ClipContextMixin.java
Rubydesic: Are you sure there's no other clip method you can use that doesn't require an entity?

copperwarrior: As far as I can tell, there's no version of clip that doesn't use a ClipContext, and ClipContext requires an entity. All this mixin does is make parity with the forge implementation of ClipContext.

Rubydesic: If there's truly no way to do this without a mixin at least add a comment explaining that

PR #184, opened 2022-11-01
Feature/commands

ewoudje:
Main pr https://github.com/ValkyrienSkies/vs-core/pull/39


Rubydesic:
If we're going to merge this into `vs-core`, we can't expose Brigadier types from `vs-core` like it currently does. That, or we need to relocate Brigadier in `vs-core` and convert the relocated the `vs-core` version of Brigadier types into the Minecraft version of Brigadier inside of this PR. At any rate, we can't rely on Minecraft and `vs-core` having the same version of Brigadier.

Personally, I would much prefer if vs-core could not expose brigadier at all - we basically just need to expose `String executeCommand(String)` and `List<String> getAutocompletions(String)` with some type of system for text colorization, unless new versions of Minecraft do something fancy that I'm unaware of.

PR #200, opened 2022-11-06
Fix camera hitting ship walls when in custom third person

copperwarrior:
When in third person mounted on a ship, third person will not hit the ship and instead use the right zoom out.
Needs testing on Forge before merge, my vs2 dev env is broken with eureka installed


copperwarrior:
ok tested it on forge it works there too

StewStrong (approved):

PR #201, opened 2022-11-07
Update to 1.18.x, Add native quilt support

AlphaMode:
This PR includes:

- Update to 1.18.2
- Native Quilt support

Currently things need more testing to find bugs but its generally playable


Rubydesic:
todo:

- [ ] 
- [ ] not working with forge, needs to jij forge dependencies because forge uses new java module system. Also JIJ doesn't respect transitive dependencies so need to consider that
- [ ] forge doesn't work in dev environment
- [ ] quilt compat is too much copy pasting. need some subproject to reduce that

ewoudje:
Ah shit did i do this accedntly

PR #210, opened 2022-11-08
Added CC: Restitched Compat - Wireless

TechTastic:
Wireless Modems and components now work for:
- Ship-to-Ship Communication when within correct distance in world space
- Ship-to-World Communication when within correct distance in world space


Bottinator22:
What about distance checks for CC's GPS feature?

copperwarrior:
This should be ported to work with the Forge version of ComputerCraft, CC: Tweaked I believe?
If the code is similar enough this should not be difficult to do.

fabric/build.gradle
```diff
@@ -24,6 +24,16 @@ loom {
     accessWidenerPath = project(":common").loom.accessWidenerPath
 }
 
+repositories {
```
Rubydesic, line 27: Remove the extra repositories before merging. Our CI should automatically pull the necessary artifacts into our maven

Rubydesic (changes requested):

PR #221, opened 2022-11-12
Changes to relocation utils

copperwarrior:
relocateBlock uses vanilla BlockState.rotate() for compatibility, and so we don't reinvent the wheel
relocateBlock has the option to update blocks around it or not
updateBlock method, which more thoroughly updates blocks copied from world to shipyard
Assembly updates blocks separately after copying all of them

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
@@ -9,7 +9,6 @@ import net.minecraft.world.level.block.Rotation
 import net.minecraft.world.level.block.Rotation.NONE
 import net.minecraft.world.level.block.state.BlockState
 import net.minecraft.world.level.chunk.LevelChunk
-import net.minecraft.world.level.chunk.LevelChunk.EntityCreationType.CHECK
 import org.valkyrienskies.core.api.ServerShip
 import org.valkyrienskies.mod.api.ShipBlockEntity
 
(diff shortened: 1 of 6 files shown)


copperwarrior:
Eureka's disassembly will need to be updated to use the new relocateBlock

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -48,70 +47,60 @@ fun relocateBlock(
...
+ * @param fromPos old position coordinate
+ * @param toPos new position coordinate
+ * @param toState new blockstate at toPos
+ */
+fun updateBlock(level: Level, fromPos: BlockPos, toPos: BlockPos, toState: BlockState) {
+
+    level.setBlocksDirty(fromPos, toState, AIR)
+    level.sendBlockUpdated(fromPos, toState, AIR, 75)
+    //This handles the update for neighboring blocks in worldspace
+    AIR.updateNeighbourShapes(level, fromPos, 75 and -0x22, 512 - 1)
```
ewoudje, line 82: what the hell is going on here?! Please add a comment that explains from where u got these numbers and wherefor they are used.

ewoudje (changes requested):
Also please run kotlin format in gradle, even though we are lately breaking format allot, i think 
 its not a bad idea to still keep it somewhat  formatted.

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
copperwarrior: Documented. Not sure where the -0x22 came from, but it was unnecessary, and so was removed.

ewoudje (approved):
Looking good, il wait until build finishes

PR #226, opened 2022-11-15
Update plants.json

Weromano:
added accurate weights to kelp and sea grass, as the blocks include a water block by default, which i think, wasnt accounted for in physics


Rubydesic:
This PR is marked as a draft, do you want it merged or still working on it?

PR #262, opened 2022-12-16
[1.16.5] Minimum ComputerCraft Compatibility Code

TechTastic:
This PR is for the minimum amount of changes necessary to fix all major incompatibilities and potential issues between VS2 and ComputerCraft (both Restitched and Tweaked)

Both have the same Mixins outside of MixinSpeakerPosition which exists exclusively on Forge

_BIg thanks to @copperwarrior for MixinSpeakerPeripheral_

- **MixinSpeakerPosition**
    - Fixes getPosition() for anywhere that may call it whilst on a ship

- **MixinSpeakerPeripheral**
    - Fixes Speakers on Ships not playing sounds in worldspace but in Shipyard

- **MixinTurtleBrain**
    - Fails Movement if the turtle is in worldspace and the target position is a
      ship (returns "ship")
    - Fails Movement if the turtle is on ship and the target position is outside
      a ship's ActiveChunkSet (returns "out of ship")

- **MixinTurtleMoveCommand**
    - Allows a Turtle to Leave Ships if the target position is outside of ship's
      VoxelAABB and ship is not scaled up or down

- **MixinWirelessNetwork**
    - Fixes Modems not being able to communicate from/to ships despite being
      within range


walksanatora:
why closed?

PR #264, opened 2022-12-17
Fixes randomTick (crops) on ships for 1.18.x

FluffyJenkins:
It looks like the logic flipped from "No-one here?" to "Anyone here?" from 1.16.x to 1.18.x but a ! was missed

I've tested this locally and it fixes crops for me.


walksanatora:
Merging this should fix issues mentioned in #264 also

Rubydesic:
thank you for the PR. Impressive what an impact a single character can have, huh?

PR #265, opened 2022-12-17
[1.18.2] Minimum ComputerCraft Compatibility Code

TechTastic:
This PR is for the minimum amount of changes necessary to fix all major incompatibilities and potential issues between VS2 and ComputerCraft (both Restitched and Tweaked)

Both have the same Mixins outside of MixinSpeakerPosition which exists exclusively on Forge

_BIg thanks to @copperwarrior for MixinSpeakerPeripheral_

- **MixinSpeakerPosition**
  - Fixes Speakers on Ships not playing sounds in worldspace but in Shipyard

- **MixinTurtleBrain**
  - Fails Movement if the turtle is in worldspace and the target position is a ship (returns "ship")
  - Fails Movement if the turtle is on ship and the target position is outsidea ship's ActiveChunkSet (returns "out of ship")

- **MixinTurtleMoveCommand**
  - Allows a Turtle to Leave Ships if the target position is outside of ship's VoxelAABB and ship is not scaled up or down

- **MixinWirelessNetwork**
  - Fixes Modems not being able to communicate from/to ships despite being within range


TechTastic:
This is finally ready to be merged

It has all been tested both Forge and Fabric, in and out of dev

fabric/build.gradle
```diff
@@ -24,6 +24,16 @@ loom {
     accessWidenerPath = project(":common").loom.accessWidenerPath
 }
 
+repositories {
+    // CC Restitched
+    maven {
+        url "https://cursemaven.com"
```
StewStrong, line 30: I think Ruby wants us to put artifacts in the VS maven. cc @Rubydesic

forge/build.gradle
```diff
@@ -50,6 +50,16 @@ configurations {
     shadowCommon.extendsFrom core
 }
 
+repositories {
+    // CC Tweaked
+    maven {
+        url "https://cursemaven.com"
```
StewStrong, line 56: Same issue as fabric

StewStrong (commented):
All lgtm, though Ruby might want to change the gradle to use the vs2 maven only

StewStrong:
We can fix the maven later, Ill just merge it

PR #268, opened 2022-12-17
superglue bandaid fix

ThePlasticPotato:
Literally just adding superglue to the list of vs shipyard entities. Does not fix latest create yet, and glue doesn't seem to want to render, but it does stick.


Rubydesic:
Thank you for your pr

PR #274, opened 2022-12-19
[1.18.2] ComputerCraft Integration

TechTastic:
Adds a universal Ship peripheral to any computer on a Ship

**ValkyrienSkiesPeripheralProviders**
- Needed to not have Forge and Fabric throw errors when they cannot find ComputerCraft classes

**ShipPeripheralProvider**
- Performs checks to ensure the Computer exists, that its on a Ship, and that the Peripheral is always in front of the computer

**ShipPeripheral**
- Adds multiple methods from Eureka's ShipHelmPeripheral to cut down on potential ComputerCraft Integration bloat (_every addon having a getMass(), getPosition(), getRotation(), etc will get quite repetitive and excessive if more than one addon is used_) at a time
- Methods Added:
  - getMass() - returns Ship's mass
  - getPosition() - returns Ship's position in worldspace as a table with X, Y, and Z
  - getRotation - return Quaternion in the form of a table with X, Y, Z, and W
  - getVelocity() - returns a Vector3 in the form of a table with X, Y, and Z
  - getScale() - returns a Vector3 in the form of a table with X, Y, Z
  - getShipID() - returns Ship's ID
  
**Changes to Existing Files**
- VSGameConfig
  - Added ComputerCraft segment to Server Config
  - Added option for disabling ComputerCraft integration
- ValkyrienSkiesModFabric
  - Tests if ComputerCraft is loaded and if it is not disabled before calling ValkyrienSkiesPeripheralProvider
- ValkyrienSkiesModForge
  - Tests if ComputerCraft is loaded and if it is not disabled before calling ValkyrienSkiesPeripheralProvider


walksanatora:
why is it always infront of the computer
why is it not a seperate block that can be placed *somewhere* on a modem network

TechTastic:
> why is it always infront of the computer
> why is it not a seperate block that can be placed *somewhere* on a modem network

Because they don't like adding more blocks to VS2 as its only meant to be a backend for addons, not for its own content

This way they don't have another block but all the functionality of the Peripheral

walksanatora:
But what if I have a block infront of the computer that is also a peripheral

walksanatora:
Because isn't there a way to say a peripheral name so instead of a direction it could be `peripheral.wrap("ship")`

TechTastic:
> But what if I have a block infront of the computer that is also a peripheral

this is why i asked about making a Mod Integrations page on the Wiki

PR #296, opened 2023-01-04
Added Create Seat entity into VS_Entities

FluffyJenkins:
Hate it when ships don't take a seat when you went through the effort of offering!

### DIFF
--- a/common/src/main/resources/data/create/vs_entities/seat.json
+++ b/common/src/main/resources/data/create/vs_entities/seat.json
@@ -1,0 +1,3 @@
+{
+  "handler": "valkyrienskies:shipyard"
+}


walksanatora:
I thought seats were blocks though
Then again it makes sense that they used a entity
Create is spaghet rough as is

ThePlasticPotato:
you cant sit on a block, they have to use entities because mojank

ThePlasticPotato:
yes I know that, i meant you cant mount a block directly so they use an entity

ThePlasticPotato:
ahh sorry i didnt mean to come off as rude dfgsdfg

ThePlasticPotato:
anyways this is grossly misusing github comments :clueless:

PR #305, opened 2023-01-07
Unfloored position components in getOnPos related parts of the entity_collsion mixin

FluffyJenkins:
This fixes the getOnPos mixins where it was doing Math.floor on position components and then +.5 to get the 'block center' before the position had been transformed into ship space


copperwarrior:
code looks good to me 👍

PR #306, opened 2023-01-08
Only move ship entities into world once they're spawned in

### LINKED ISSUE #303: Ship Rotation Transforms Broke:tm:
When using Valkyrien Skies 1.18.2 Beta 8+, regardless of other mods, there is an issue with entity transforms onboard vessels.

Projectiles fired by anything within the shipyard will not properly enter realworld space- rather, they will fire in the direction the block was originally facing in the shipyard. This applies to dispensers, it applies to Create: Big Cannons, and all other forms of projectile weapon I've tested.

Presumed issue: Rotation transforms are simply not being applied to a projectile leaving the shipyard after being fired by a block or entity from within the shipyard.

How to reproduce: place a dispenser, create a vessel, rotate it, and attempt to fire. Note how the arrows do not eject in the correct direction.

Rubydesic:
setPosRaw might be called before the velocity of the entity is updated, causing the velocity of entities that were spawned in the shipyard and teleported to the world to remain untransformed.

If it turns out to that entities are often teleported into the shipyard AFTER being spawned we may want to teleport all entities out of the shipyard at the end of the tick. 

But for now, let's see if this works.

Closes #303


ewoudje (approved):
Seems good

PR #340, opened 2023-01-14
Assembly changes

copperwarrior:
Assembly changes from 1.16.5 main required for Eureka disassembly to work properly.

Tested on Forge, needs testing on Fabric.


copperwarrior:
Tested it on fabric, seems to work properly

ewoudje (approved):
Seems good to me, this is also already tested code in 1.16 so il merge it

PR #370, opened 2023-01-21
feat: add a heuristic to put seat entities in shipyard

Rubydesic:


ewoudje (approved):
Seems good, can't imagine it wrongly taking a entity to shipyard for this.

PR #404, opened 2023-02-02
Added compatibility for Create Contraption passenger positioning

FluffyJenkins:
Gradle changes to allow create imports as well, but I don't know if they break other things
PR is also for others to see the mess that is gradle


Rubydesic:
Not including this for now because the gradle changes are just way too suspicious. Ideally I'd like a way for this to be specified for any mod and not just hardcoded to create. Maybe something like an `IPassengerPositionProvider` interface if getting the passenger offset info out of the game itself is not possible?

PR #406, opened 2023-02-04
Unnecessary log print outcommented in vs-core

M-Marvin:
Removed that absolutely unnecessary debug message of receiving an network package. It spams the logs with 3 messages every tick and does not provide any information's to the user ...
It also makes the logs nearly unusable when debugging in an IDE.


M-Marvin:
I noticed that there seems to be a problem because the file "NetworkChannel.kt" is part of the vs-core "submodule".
GitHub Desktop automatically redirected mit to this and suggested to create a pull request for Valkyrien Skies 2.
And im unable to create a pull request for vs-core too.
So if something is wrong with that pull request, i just changed the line

NetworkChannel.kt Line 67
From:  logger.debug("Client received packet of type: ${packet.type}")
To:      //  logger.debug("Client received packet of type: ${packet.type}")

Rubydesic:
You need to make this pull request to the vs-core repository

Rubydesic:
Also, I would probably change it to a trace rather than commenting it out entirely

M-Marvin:
What do you mean with changing it to an trace ?
I don't know anything over Kotlin and these "submodules".
I just got annoyed by this log-spam.
GitHub does not let me make any commits to vs-core, no idea why.
All options of committing or making a pull request are grayed out.
I will try it again if i have time.

walksanatora:
a submodule is a link to another git repo
so you should make the pr in the [VS-Core](https://github.com/ValkyrienSkies/vs-core) repository
@M-Marvin

PR #411, opened 2023-02-06
1.18.x tis-3d rendering+infared fixes

walksanatora:
This fixes rendering issues on TIS-3D
meaning that TIS-3D can actually be used on ValkrienSkies ships now (AFAIK)


Rubydesic:
Have you tested this compatibility outside of your dev environment? (by building the mod as a JAR and using it normally with TIS-3D)

Rubydesic:
I see some suspicious remap=false

walksanatora:
I was told to remap=false when targeting non-mc classes

Rubydesic:
Did you test it?

walksanatora:
okay after a quick test
rendering: fixed
infared: not fixed (does not work ship within ship, ship <-> world and ship <-> other ship are ValkComp things)
i need to figure out why infared is like this, but this makes TIS-3d atleast *usable* on ships (although not 100%)

walksanatora:
so inorder to fix the Infared i gotta fix a raycast
do you know if VSGameUtilsKt has a function to replace the raycast

walksanatora:
infared still broken, when i get home gonna run under a debugger (darn forgeGradle runClient not working with --offline)

walksanatora:
@Rubydesic Finally works 
i tested and it works
the only reason why infared packets can miss is because you were going too fast or rotating too quickly (since they are pretty slow entities) but otherwise it works perfectly (even on rotated ships)

PR #413, opened 2023-02-08
ShipSettings

ewoudje:
Rn, i want this for the addon islands. But can be handy for other addons too ofc; 

Made a new attachment called ShipSettings wich controls certain behavior of mixins on a per ship basis. Rn it only controls world gen but it prob can have more usecases.

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinChunkMap.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinChunkMap.java
@@ -25,9 +25,11 @@ import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.valkyrienskies.core.api.ships.ServerShip;
 import org.valkyrienskies.core.apigame.world.IPlayer;
 import org.valkyrienskies.mod.common.VSGameUtilsKt;
 import org.valkyrienskies.mod.common.util.MinecraftPlayer;
+import org.valkyrienskies.mod.common.util.ShipSettingsKt;
 
 @Mixin(ChunkMap.class)
 public abstract class MixinChunkMap {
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinChunkMap.java
```diff
@@ -56,8 +58,10 @@ private void preReadChunk(final ChunkPos chunkPos, final CallbackInfoReturnable<
         final CompoundTag originalToReturn = compoundTag == null ? null :
             self.upgradeChunkTag(this.level.dimension(), this.overworldDataStorage, compoundTag, Optional.empty());
 
+        cir.setReturnValue(originalToReturn);
         if (originalToReturn == null) {
-            if (VSGameUtilsKt.isChunkInShipyard(level, chunkPos.x, chunkPos.z)) {
+            final ServerShip ship = VSGameUtilsKt.getShipManagingPos(level, chunkPos.x, chunkPos.z);
+            if (ship != null && !ShipSettingsKt.getSettings(ship).getShouldGenerateChunks()) {
```
Rubydesic, line 64: can we make this an `if ((ship == null && VSGameUtilsKt.isChunkInShipyard(level, chunkPos.x, chunkPos.z)) || (ship != null && !ShipSettingsKt.getSettings(ship).getShouldGenerateChunks())` just in case something bad happens and someone tries to generate chunks in the shipyard where no ship exists we can prevent it

common/src/main/kotlin/org/valkyrienskies/mod/common/util/ShipSettings.kt
```diff
@@ -0,0 +1,14 @@
+package org.valkyrienskies.mod.common.util
+
+import org.valkyrienskies.core.api.ships.ServerShip
+import org.valkyrienskies.core.api.ships.getAttachment
+import org.valkyrienskies.core.api.ships.saveAttachment
+
+data class ShipSettings(
+    // Should the ship chunks try to generate? If true it will generate chunks in the shipyard.
```
Rubydesic, line 8: please use javadoc comment style /** */

Rubydesic (changes requested):

common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinChunkMap.java
ewoudje: Oh ic i didn't know that there was a difference in isChunkInShipYard and getManagingShip

Rubydesic:
lgtm

PR #421, opened 2023-02-18
Simplify type of TE

crifurch:
Small fix of type, I am found few mods that don't regard Container implementation, to improve level of compat prefer to use type that contains target method


crifurch:
If you will approve that PR, i think it should be cherry-picked to other versions

crifurch:
I fixed bug with dupe items in containers that not implements Clearable and that create ship broke machines from create(maybe in some another mods), but stuck on issue with belt that wont work correctly after relocation. And are here someone with who i can contact about dev rules and other questions?

Rubydesic:
Thank you very much for the PR, lgtm

PR #432, opened 2023-03-12
Fixing performance #431

### LINKED ISSUE #431: Significant impact on tickrate even with no ships present
I've been profiling my modpack to find what makes the TPS go down and found, not without some help, that some of the VS2 mixins cause rather significant impact.

Namely, the postTick() mixins:
https://github.com/ValkyrienSkies/Valkyrien-Skies-2/blob/426ce0d5815c8700b2cc843b638b921905dffbc7/common/src/main/java/org/valkyrienskies/mod/mixin/server/MixinMinecraftServer.java#L177
and
https://github.com/ValkyrienSkies/Valkyrien-Skies-2/blob/d84531555f6d27171b763ffb6d6c9575defe3b36/common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinServerLevel.java#L118

that seem to be taking up around 18% of the server tick time.
A large part of this seems to be caused by the HashSet.add() method which seems to likely be currentTickChunkRegions.add(chunkPos);, but I'm not entirely sure.

Here's my Spark profile result in case it provides any additional insight:
https://spark.lucko.me/3rK7FKNaCE

And here is a screenshot of the part that I'm referring to:
image

Hope you have the time to take a look at this and perhaps optimise it somehow.
The test world doesn't have a single ship in it, yet.

ewoudje:
Untested rn, forge repo was sussin. After using vs2 repo gradle went sussin so i didn't test it.

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinServerLevel.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/server/world/MixinServerLevel.java
@@ -7,7 +7,6 @@ import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


ewoudje:
Tested it, all old ships loaded in and new ships work. If something was wrong with it prob wouldve been *quite* visible.

PR #457, opened 2023-04-27
Feature/commands 2

StewStrong:


common/src/main/kotlin/org/valkyrienskies/mod/common/command/VSCommands.kt
```diff
@@ -0,0 +1,166 @@
...
+                                            } catch (e: Exception) {
+                                                if (e !is CommandRuntimeException) LOGGER.throwing(e)
+                                                throw e
+                                            }
+                                        })
+                            )
+
+                            // Scale a ship
+                            .then(
+                                literal("scale")
```
ewoudje, line 90: Since this doesn't work maybe comment it out, so that ppl aren't like 'why isn't this working!!!'

ewoudje (approved):

PR #491, opened 2023-06-09
Change Sodium maven source to jitpack

littlej541:
Currently, Curse Maven is returning 403 for Sodium. This may only be related to the recent malware troubles, but Sodium is no longer uploading new versions to CurseForge, so it may be worth changing the source anyways.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,6 +89,7 @@ subprojects {
             maven { url = "https://maven.tterrag.com/" } // Registrate, Forge Create and Flywheel
             maven { url = "https://maven.cafeteria.dev/releases" } // Fake Player API
             maven { url = "https://maven.jamieswhiteshirt.com/libs-release" } // Reach Entity Attributes
+            maven { url 'https://jitpack.io' }
         }
     }
 
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -8,7 +8,7 @@ dependencies {
 
     modApi("me.shedaniel.cloth:cloth-config:4.14.64")
 
-    modCompileOnly("curse.maven:sodium-394468:3669187")
+    modCompileOnly("com.github.CaffeineMC:sodium-fabric:mc1.18.2-0.4.1")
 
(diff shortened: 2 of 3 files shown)


Rubydesic:
suggest you set `block_external_repositories=true` in the `build.gradle` if you are having trouble with external maven servers. jitpack also frequently has problems so I am closing this issue.

littlej541:
My issue was actually Eureka failing because its VS2 dependency couldn't grab Sodium. I'm not that experienced with gradle, so how would I go about fixing that? I just built VS2 locally as a workaround.

Rubydesic:
Sorry, I mean gradle.properties, not build.gradle

Rubydesic:
keep in mind that if you are building eureka, only repositories in eureka will be resolved, not repositories in the gradle of valkyrien skies

littlej541:
Then I'm not sure why Eureka was failing with `Could not GET 'https://cursemaven.com/curse/maven/sodium-394468/3669187/sodium-394468-3669187.jar'. Received status code 403 from server: Forbidden`. The only place I could find that was in VS2. :shrug:

Ah well, what I'm doing is working.

littlej541:
So just a heads-up in case you were interested:

I did a little bit of digging and found that Sodium _is_ included in Eureka via the VS2 POM files. If you look at [118-fabric-2.1.0-beta.10+f520a5c13d.pom](https://maven.valkyrienskies.org/org/valkyrienskies/valkyrienskies-118-fabric/2.1.0-beta.10+f520a5c13d/valkyrienskies-118-fabric-2.1.0-beta.10+f520a5c13d.pom), you'll find a dependency entry for it on lines 309-314. So I couldn't build Eureka because Sodium wasn't resolving since CurseForge was returning 403 for the archived Sodium jar (likely because of the malware thing, unarchived jars were still resolving) grabbed via Curse Maven.

In any case, it's not returning 403 anymore so I suppose the issue is now moot.

PR #523, opened 2023-07-02
1.18.x / Dynamic Connectivity

ThePlasticPotato:
Dynamic Connectivity and ship splitting/welding for 1.18!

- Weld tool for welding
- Ship split detection

The code isn't perfect and is a little messy, so advice on tweaks is appreciated

also i did change it to use my fork of vscore, unsure if I need to do anything to fix that before its merged


omgcchheeessee:
let us download 🤑🤑🤑🤑

PR #529, opened 2023-07-13
Mass Tooltip for All BlockItems

TechTastic:
Utilizing a Mixin, I have added a tooltip to all BlockItems to display the masses of their associated blocks

This tooltip's visibililty is changeable via the client config where it can be set to always on, only when advanced tooltips are enabled, or disabled

There is also a config option to use Imperial Units when displaying mass

I also added a translation key for "Mass"


StewStrong:
Very cool :)

PR #534, opened 2023-07-14
Ported Mass Tooltip from 1.18.2

TechTastic:
As the title says, this is a port of my Mass Tooltips to 1.19.2


StewStrong:
This was ported in merge commits to 1.19 and 1.20

PR #572, opened 2023-09-04
biome foilage

SuperCraftAlex:
Biome foilage and similar working with VS2 ships


Rubydesic:
Does this work if the ship moves biomes without rebuilding the chunk vbos?

SuperCraftAlex:
> Does this work if the ship moves biomes without rebuilding the chunk vbos?

no

ewoudje:
Hmm I don't know what is best to have then, or it stays the same forever or you get jumpscared on relogging.
But rebuilding vbo's sounds like no solution..

ewoudje (approved):
LGTM Im happy its disabled by default, since the behavior is pretty specific.

common/src/main/java/org/valkyrienskies/mod/mixin/feature/block_tint/MixinClientLevel.java
```diff
@@ -0,0 +1,68 @@
...
+        "tintCaches"
+    )
+    public abstract Object2ObjectArrayMap<ColorResolver, BlockTintCache> getTintCaches();
+
+    /**
+     * @author SuperCraftAlex
+     * @reason Fix block tinting for ships
+     */
+    @Overwrite
+    public int getBlockTint(BlockPos blockPos, ColorResolver colorResolver) {
```
Rubydesic, line 57: I can almost guarantee this overwrite is going to cause compatibility issues with other mods. Can we please rewrite this in a more compatible way?

Rubydesic: cc @ewoudje @SuperCraftAlex

ewoudje: You think mods will modify block tint? I was fine with it since I thought it wouldn't matter.

ewoudje: But I suppose a inject at head with a recall would be better

ewoudje: Ah was already fixed 👍

PR #596, opened 2023-10-05
ship-related render events

alex-s168:
Adds a modifiable ship render event that allows client-side transformation of ship positions and a Shipyard entity render transform event

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/vanilla_renderer/MixinLevelRendererVanilla.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/vanilla_renderer/MixinLevelRendererVanilla.java
@@ -15,7 +15,6 @@ import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import it.unimi.dsi.fastutil.objects.ObjectList;
 import java.util.ListIterator;
 import java.util.WeakHashMap;
-import java.util.concurrent.atomic.AtomicReference;
 import net.minecraft.client.Minecraft;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/vanilla_renderer/MixinLevelRendererVanilla.java
```diff
@@ -178,12 +179,25 @@ private void redirectRenderChunkLayer(final LevelRenderer receiver,
                 center.x(), center.y(), center.z(),
                 camX, camY, camZ);
 
+            final AtomicReference<Matrix4f> matrix4fNew = new AtomicReference<>(matrix4f);
```
StewStrong, line 182: Why are these atomic? Atomic is generally slow so unless this needs to be run on multiple threads we should use something else.

StewStrong: Just write something like `class MutableObject<T> { var object: T }`. Its effectively a modifiable reference but it won't cause lag.

Rubydesic:
Can I see a practical application of this event? It seems like physics mod wanted to update the render transform entirely (which we will be adding support for). We can include this event in the new API (see #608), but there are a lot of parameters here and its not clear which ones are necessary or how this would be used.

ewoudje:
Id like to close this in favor of https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/608
This alright for you @alex-s168

PR #607, opened 2023-10-22
1.18.x/feature/dynconn

ThePlasticPotato:
funny dynnconn real


ThePlasticPotato:
This is invalidated and completely rendered useless by the new vscore side of dynconn so 😇

PR #625, opened 2023-11-16
1.20.1/update

mrsterner:
Still need to do org.valkyrienskies.mod.mixin.server.MixinMinecraftServer, since the portal stuff got very much changed, also need to init items


mrsterner:
There we go, everything seems to work except portals

PR #629, opened 2023-11-19
Modular Routers Compatibility

TechTastic:
Added a Mixin to fix the Modular Router UI distance check when on ships


StewStrong:
Thank you, very epic! :DDDD

PR #648, opened 2023-12-11
Position based gravity among other things

DatCoder464:


DatCoder464:
I plan on using the current features for Malumian Skies and Universal
I did 2 things 
  1. I added a more mainstream way to access the camera
  2. I made a gravity ship controller that also changes the position of the position (as well as there camera) which can be used by 

calling the .PostionGravity(postion, mass) (im chaning it to .addGravity(postion, mass)) and removing it with .removeGravity(position) from the PostionGravity class
its needed for another PR I'm working on for centrifugal forces as well as VSFX and Universal

common/src/main/java/org/valkyrienskies/mod/mixin/client/MixinCamera.java
```diff
@@ -128,6 +128,19 @@ private void setRotationWithShipTransform(final float yaw, final float pitch, fi
         this.left.transform(this.rotation);
     }
 
+    @Unique
+    public void addRotation(final Quaterniondc quaterniondc) {
```
StewStrong, line 132: You should not add public functions to mixins, unless you are implementing an interface.

PR #651, opened 2023-12-16
added a null check and fixed missing parenthesis

mrsterner:
this "fixes" crashed on world reload, might be a better solution so feel free to close pr and make a better one


common/src/main/java/org/valkyrienskies/mod/mixin/client/MixinMinecraft.java
```diff
@@ -159,4 +159,14 @@ public void deleteShipObjectWorldClient() {
         shipObjectWorldCopy.destroyWorld();
         shipObjectWorld = null;
     }
+
+    @Inject(
+        method = "clearLevel",
+        at = @At("TAIL")
```
StewStrong, line 165: Should this be @At("RETURN") instead?

mrsterner: Method doesn't return anything so it basically synonymous

StewStrong (approved):

PR #690, opened 2024-01-12
Fix Create Schematic placement, rendering on ships

Endalion:
- Fix the really weird position for schematic/overlay to work on ships
- Mix schematic renderer to (also) work on ships

Initial solution very ugly open to feedback


Rubydesic:
Could you just add some comments/javadocs explaining what these mixins do? Thanks

common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/create/client/MixinDeployTool.java
```diff
@@ -0,0 +1,81 @@
...
+            VSClientGameUtils.transformRenderWithShip(ship.getTransform(), ms, x - centerX, y, z - centerZ, camera.x, camera.y, camera.z);
+        } else {
+            ms.translate(x - centerX - camera.x, y - camera.y, z - centerZ - camera.z);
+        }
+    }
+
+    // To fufill the class extension requirements
+    @Unique
+    @Override
+    public boolean handleRightClick() {
```
ewoudje, line 72: Make your class abstract, this is sus idk how this would behave with other mixins

ewoudje (changes requested):
looks good except the 2 methods overriding there

common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/create/client/MixinDeployTool.java
StewStrong: abstract pls

Endalion: I have learnt many things in these 6 long months

PR #692, opened 2024-01-17
Create Compat(1.18.x): Fixed track outlines and picking

FluffyJenkins:
Fixes:

- track block/curve outlines render when you're on a ship
- track blocks/curves will actually highlight/select when you're not looking at its center block like it should
- Having a ship within 10m of you with curved track around will no longer pick/highlight the curve instead of what you're actually looking at


ewoudje (approved):
LGM, altough another name for the mixin then 'stuff' would be more optimal butt ig its pretty clear what it does

PR #736, opened 2024-02-16
Create Compat(1.19.2): Removed old file causing crashes/renamed function

FluffyJenkins:
Forgot to delete old MixinTrackBlockOutline and so it caused crashes....
Don't know how I missed this when it wasn't causing a crash during testing!


ewoudje:
LGTM

PR #743, opened 2024-02-20
Fixed patch f crash and added some support for Create utilities

mrsterner:
-Added support for VoidLinkBehaviour
-Fixed MixinTrackBlockOutline


ewoudje:
Can there be some java docs for the mixins? explain in the docs why this mixin is needed and what it does

mrsterner:
@ewoudje

Issue #818, opened 2024-04-18
Crash when entering the game, with Sodium and Litematica mods


pama1234:
### Mod Name

Sodium, Litematica

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [X] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.20.x

### Mod Loader

Fabric

### Issue description

The Game crashed.

related issues：
- https://github.com/maruohon/litematica/issues/501

the problem: 
- https://github.com/ValkyrienSkies/Valkyrien-Skies-2/blob/d9e0db7d3354dcd03c4a8a40c3d72ed3ee664cba/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/sodium/MixinSodiumWorldRenderer.java#L53

### Issue reproduction

install Sodium, Litematica(and MaLiLib), and Valkyrien-Skies on a 1.20.1 fabric client, create and enter a single player map.

### Logs

java.lang.ClassCastException: class fi.dy.masa.litematica.world.WorldSchematic cannot be cast to class net.minecraft.class_638 (fi.dy.masa.litematica.world.WorldSchematic and net.minecraft.class_638 are in unnamed module of loader net.fabricmc.loader.impl.launch.knot.KnotClassLoader @103f852)
	at me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer.redirect$hhg000$valkyrienskies$renderShipBlockEntityInShipyard(SodiumWorldRenderer.java:1572) ~[sodium-fabric-0.5.8+mc1.20.1.jar:?]
	at me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer.renderBlockEntity(SodiumWorldRenderer.java:349) ~[sodium-fabric-0.5.8+mc1.20.1.jar:?]
	at me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer.renderBlockEntities(SodiumWorldRenderer.java:308) ~[sodium-fabric-0.5.8+mc1.20.1.jar:?]
	at me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer.redirect$hhg000$valkyrienskies$renderShipBlockEntities(SodiumWorldRenderer.java:1598) ~[sodium-fabric-0.5.8+mc1.20.1.jar:?]
[...]

PR #819, opened 2024-04-18
fix crash with Sodium and Litematica

pama1234:
fix https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/818

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/sodium/MixinSodiumWorldRenderer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/sodium/MixinSodiumWorldRenderer.java
@@ -50,6 +50,10 @@ public abstract class MixinSodiumWorldRenderer {
         final double camZ, final BlockEntityRenderDispatcher dispatcher, final BlockEntity entity) {
 
         final BlockPos pos = entity.getBlockPos();
+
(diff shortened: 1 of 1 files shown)


Hex4dec:
You should probably do the client check way earlier, maybe at the start of each render function. And why are you doing the client check through dispatcher? Just do level.isClient?

ewoudje (approved):
LGTM

dti131:
image
Fix #819 does fix the crashing issue, but it adds another issue (much less severe) where blocks that have animations or other entities (such as the blaze burner and conveyor belt) will be rendered near the player. This can be "fixed" by moving to an area without those blocks, though.

chen-0917:
> image修复 #819 确实修复了崩溃问题，但它增加了另一个问题（不那么严重），即具有动画或其他实体（例如烈焰燃烧器和传送带）的块将在玩家附近渲染。不过，这可以通过移动到没有这些方块的区域来“修复”。
Can you send me the built file, I'm a novice, and I tried for a day but I kept getting an error

Ozzer0:
> image Fix #819 does fix the crashing issue, but it adds another issue (much less severe) where blocks that have animations or other entities (such as the blaze burner and conveyor belt) will be rendered near the player. This can be "fixed" by moving to an area without those blocks, though.

Having a similar issue with vanilla block entities.

PR #827, opened 2024-04-29
Create ko_kr.json

Merhaf:
Korean localization add


ewoudje:
LGTM not that i know whats written there

Issue #832, opened 2024-05-06
TFC compatibility mixin error


Dokueki:
### Mod Name

TerraFirmaCraft

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [X] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

What happens: Throwing error during game launch (preGetBaseColumn in MixinTFCChunkGenerator missing RandomState argument)
Expected behaviour: Game launches normally

### Issue reproduction

1. Install both Valkyrien Skies and TerraFirmaCraft
2. Launch the game

### Logs

[debug.log](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/files/15219846/debug.log)
[latest.log](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/files/15219847/latest.log)

thetheaplant:
Fixed in #833

PR #833, opened 2024-05-06
Fix TFC compatibility issue

Dokueki:
https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/832

### DIFF
--- a/forge/src/main/java/org/valkyrienskies/mod/forge/mixin/compat/tfc/MixinTFCChunkGenerator.java
+++ b/forge/src/main/java/org/valkyrienskies/mod/forge/mixin/compat/tfc/MixinTFCChunkGenerator.java
@@ -59,7 +59,7 @@ public class MixinTFCChunkGenerator {
     }
 
     @Inject(method = "m_214184_", at = @At("HEAD"), cancellable = true)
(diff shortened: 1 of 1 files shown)


MayTheCutie:
This isn't fixed. Y'all might need to change where the mixins of world creation are, since you're referencing vanilla chunkgen. Would y'all be able to figure out a different way for this since there's other mods that have chunkgen?

Rubydesic:
> This isn't fixed. Y'all might need to change where the mixins of world creation are, since you're referencing vanilla chunkgen. Would y'all be able to figure out a different way for this since there's other mods that have chunkgen?
> 
> It'd help the mod more if you made this a tad more compatible to other mods too, not just tfc

Definitely! I'm more than happy to review a PR if you have a better way of managing shipyard chunk generation in mind. Merry Christmas!

IslamSir:
I think will be better to change the TFC chunkgen, if you say so

ISGRain:
seams relatively compatible in 1.18 but not in 1.20, not sure if it just wasn't implemented in both or if tfc changed its gen since figured it was worth bringing up though

PriestOfFerns:
@ISGRain  Did you test using the github versions of VS2?

PR #854, opened 2024-05-27
Replaced assembly code by M_Marvin (Ported to Kotlin/Multiloader by PriestOfFern)

ThePlasticPotato:
Self Explanatory

Might need some review

Didn't want to remove the old assembly functions as to not break existing addons so I deprecated them for now

Should also be able to handle assembling ships out of ships supposedly, unable to test rn while I redo splitting

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/AssemblyUtil.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/AssemblyUtil.kt
@@ -1,0 +1,148 @@
+package org.valkyrienskies.mod.common.assembly
+
+
+import com.mojang.math.Vector3d
+import net.minecraft.core.BlockPos
+import net.minecraft.core.Direction
+import net.minecraft.nbt.CompoundTag
+import net.minecraft.util.Mth
+import net.minecraft.world.level.Level
+import net.minecraft.world.level.block.Block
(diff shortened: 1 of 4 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/AssemblyUtil.kt
```diff
@@ -0,0 +1,147 @@
...
+            val data: CompoundTag = blockentity.saveWithId()
+            level.setBlockEntity(blockentity)
+            val newBlockentity = level.getBlockEntity(to)
+            newBlockentity?.load(data)
+        }
+    }
+
+    fun updateBlock(level: Level, fromPos: BlockPos, toPos: BlockPos, toState: BlockState) {
+
+        // 75 = flag 1 (block update) & flag 2 (send to clients) + flag 8 (force rerenders)
```
ewoudje, line 53: Why does it say 75 =, when its 11 and also flag 8 has a + instead of a &

PriestOfFerns: ewou, this PR is outdated. Tomato added the assembly code while they were adding vs-api to 1.18

ewoudje:
Close it then...

PriestOfFerns:
@ewoudje do I look like I have PR perms

PR #897, opened 2024-07-09
Weather 2 wind & storm affecting ships

alex-s168:
(identical to #896 but removed the other git author (which was me))

The values in the config should be adjusted a bit

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/server/MixinMinecraftServer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/server/MixinMinecraftServer.java
@@ -62,6 +62,7 @@ import org.valkyrienskies.mod.common.util.VSLevelChunk;
 import org.valkyrienskies.mod.common.util.VSServerLevel;
 import org.valkyrienskies.mod.common.util.VectorConversionsMCKt;
 import org.valkyrienskies.mod.common.world.ChunkManagement;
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/server/MixinMinecraftServer.java
```diff
@@ -229,6 +230,7 @@ private void postTick(final CallbackInfo ci) {
         // Only drag entities after we have updated the ship positions
         for (final ServerLevel level : getAllLevels()) {
             EntityDragger.INSTANCE.dragEntitiesWithShips(level.getAllEntities());
+            Weather2Compat.INSTANCE.tick(level);
```
StewStrong, line 233: Can you add an if statement here to only run Weather2Compat if the Weather2 mod is loaded?

StewStrong:
Can you modify the code to only run if Weather2 is installed? Right now it seems like this will crash if someone uses VS2 without Weather2.

alex-s168:
There is a `if (!LoadedMods.weather2) return` in the function
I'm pretty sure that the jvm only loads classes (in this case weather2) when it is used in the code

alex-s168:
@StewStrong moved it anyways because nicer code

StewStrong:
Nice! :D

PR #907, opened 2024-07-18
Minor fix, Urgent: Fix soft-overwrite preventing other mods' mobs from despawning with special conditions.

ByThePowerOfScience:
Replaces a soft-overwrite Inject with a ModifyArg that does the same thing and doesn't prevent other Mixins targeting this method from firing.

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/shipyard_entities/MixinPersistentEntitySectionManager.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/shipyard_entities/MixinPersistentEntitySectionManager.java
@@ -61,7 +61,7 @@ public abstract class MixinPersistentEntitySectionManager implements OfLevel {
 					value="INVOKE"
 			)
 	)
-	private LongPredicate foo(LongPredicate par1) {
+	private LongPredicate processUnloads_catchException(LongPredicate par1) {
 		return (l) -> {
 			try {
 				return par1.test(l);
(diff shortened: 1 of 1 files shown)


StewStrong:
This code fails when I try running it

ByThePowerOfScience:
> This code crashes when I try running it

I'll take a look, it's possible I got the args wrong.

StewStrong:
I think this would work, tho I havent tested it

```
package org.valkyrienskies.mod.mixin.feature.shipyard_entities;

import net.minecraft.world.entity.Entity;
import net.minecraft.world.level.Level;
import net.minecraft.world.level.entity.EntitySectionStorage;
import net.minecraft.world.level.entity.PersistentEntitySectionManager;
import org.spongepowered.asm.mixin.Final;
import org.spongepowered.asm.mixin.Mixin;
import org.spongepowered.asm.mixin.Shadow;
import org.spongepowered.asm.mixin.Unique;
import org.spongepowered.asm.mixin.injection.At;
import org.spongepowered.asm.mixin.injection.ModifyArg;
import org.valkyrienskies.mod.mixinducks.world.OfLevel;

@Mixin(PersistentEntitySectionManager.class)
public abstract class MixinPersistentEntitySectionManager implements OfLevel {
    @Shadow
    @Final
    EntitySectionStorage<Entity> sectionStorage;

    @Unique
    private Level valkyrienskies$level;

    @Override
    public Level getLevel() {
        return valkyrienskies$level;
    }

    @Override
    public void setLevel(final Level level) {
        this.valkyrienskies$level = level;
        ((OfLevel) this.sectionStorage).setLevel(level);
    }

    /**
     * This fixes this function randomly crashing. I'm not sure why but the removeIf() function is buggy
     */
    @ModifyArg(
        method = "processUnloads",
        at = @At(
            target = "Lit/unimi/dsi/fastutil/longs/LongSet;removeIf(Ljava/uti l/function/LongPredicate;)Z",
            value = "INVOKE"
        )
    )
    private java.util.function.LongPredicate processUnloads_catchException(
        final java.util.function.LongPredicate par1) {
        return (l) -> {
            try {
                return par1.test(l);
            } catch (final Exception e) {
                e.printStackTrace();
                return false;
            }
        };
    }
}
```

StewStrong:
I merged the updated fix in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/dcba513e6d0e8a49903a1b0be5dfb6baedb321f7

ByThePowerOfScience:
Nevermind, I misunderstood the problem `removeIf` was having. The function _itself_ is running into a NPE, not something in the predicate.

In that case, do what SmoothChunkLoading does and mix into `tick`'s call of `processUnloads` and wrap it with a WrapOperation handler.  I'll do it in the morning.

PR #908, opened 2024-07-20
Added missing parameter newDimension to ShipTeleportDataImpl constructor

xiewuzhiying:
Added missing parameter newDimension to ShipTeleportDataImpl constructor


Rubydesic:
lgtm

plasmadome:
So how do you do this?
How do you use the command?

BlueStevie64:
> So how do you do this? How do you use the command?

I also have been trying to figure it out and I can't seem to use the teleport command past the coordinates. the [<euler-angles>] doesn't seem to be working as whatever I put there just brings up an error called "incomplete (expected 2 coordinates)"

zyxkad:
nope it has nothing to deal with the teleport command, you need implement cross dimension teleport by yourself

PR #915, opened 2024-07-22
move mass unit conversion into util and add tooltip event

alex-s168:
title

I did this because I need to display **mass values** in **item tooltips** in Tournament and other addons might want to do similar things

I'm not really happy with the mass tooltips ("old" feature) beeing in the main mod class, but I don't know where to move the code to


ewoudje:
Could you move the mixin in its own feature, and instead of having a even abstraction just doing it directly.
Im assuming the tooltips event is for tournament use too but VS2 isn't an util library for anything except physics.

TechTastic:
Why Mixin into Item and not BlockItem, especially is you are checking for BlockItem anyways?

PR #937, opened 2024-08-07
Fix #486 - when you log off on a ship and it moves, you should log on still in the ship, not in the middle of nowhere

### LINKED ISSUE #486: when you log off on a ship and it moves, you should log on still in the ship, not in the middle of nowhere

m-doescode:
Fixes #486 

When you log off on top of a ship, or in mid-air within 4 ticks of jumping off a ship (can be changed), the id of the last ship you stood on, as well as your position in the ship's chunk claim is saved as nbt. When you reconnect, and you have the `LastShipId` tag, and it points to a valid ship, then the chunk claim position is translated to the world position and you are teleported to that position.

Please let me know of any errors/considerations. Thank you.

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
@@ -1,0 +1,89 @@
+package org.valkyrienskies.mod.mixin.feature.teleport_reconnected_player_to_ship;
+
+import com.mojang.authlib.GameProfile;
+import javax.annotation.Nullable;
+import net.minecraft.core.BlockPos;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
+import org.joml.Vector3d;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
```diff
@@ -0,0 +1,89 @@
...
+@Mixin(ServerPlayer.class)
+public abstract class MixinServerPlayer extends Player {
+
+    @Shadow
+    public abstract ServerLevel serverLevel();
+
+    // The maximum number of ticks before the player is no longer considered aboard a ship
+    // if they are no longer directly touching it
+    @Unique
+    private static final int MAX_ONBOARD_TICKS = 4;
```
ewoudje, line 32: Could you move this to the config class, so that users themselves can modify this.

ewoudje (changes requested):
LGTM except the small move to the config. Thanks for the PR!

common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
m-doescode: No problem!!

Fixed in 266c308496ecd02b4417c286ead432f648b07bcb

ewoudje (approved):
Thanks allot!

Issue #688, opened 2024-01-09
Create filtered chutes and funnels don't detect stationary items.


WhyKickAmooCow:
### Mod Name

Create

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [X] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.18

### Mod Loader

Forge

### Issue description

When bulk processing without a belt (fan jut pushing items), when an item not allowed to pass through is pushed against a filtered input and is then turned into an allowed input, it does not get picked up by the funnel/chute. When the allowed item is picked up, and thrown into the funnel it is picked up.

![2024-01-09_18 27 16](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/assets/13505247/48e4f4cb-5499-481a-9822-91b0a3690e99)
![2024-01-09_18 27 31](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/assets/13505247/b11ee160-e9a5-439d-9b18-e4b7ec27bf07)
![2024-01-09_18 27 53](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/assets/13505247/9cc356c8-a37d-4e4c-923c-7184e0f8ff83)

This appears to be the same issue reported in Creators-of-Create/Create/issues/5081

### Issue reproduction

1. Create bulk processing setup, using encased fans to push the items to the destination which is a filtered funnel or chute.
2. Set a filter on the destination filter.
3. Put recipe resource into the setup.
![2024-01-09_18 27 16](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/assets/13505247/48e4f4cb-5499-481a-9822-91b0a3690e99)

I have provided a creative world with only this setup inside it if people wanted to use it for testing.

[Bug Reproduction World.zip](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/files/13869414/Bug.Reproduction.World.zip)

TheHaskBit:
same in 1.20.1

Theuntextured:
Any updates on this?

Theuntextured:
poke. :(

Kierra1:
Same.

Zardexd:
Still happens 1.20.1 Fabric,  valkyrienskies-120-2.3.0-beta

Risync:
##**Temp solution**: 

[...]

PR #955, opened 2024-09-04
fix ItemEntity affected by Create's AirCurrent cannot be pick by Create Funnels, Fix #867, #688

### LINKED ISSUE #867: filtered funnels do not work when there is a fan affecting the item

### Minecraft Version

1.20.x

### Mod Loader

Fabric

### Issue description

i have been notified that this was a known issue with VS2.

### Issue reproduction

use VS2 as well as any farm that involves the use of the encased fan.

### Logs

_No response_

MoePus:
Create's Funnels are using `entityInside` to pick up items,
but Entity::tryCheckInsideBlocks is skipped by vs2's mixin


ewoudje (approved):

ewoudje:
Yep nice catch, Im kinda worried if we the mixin also skips other things such as checking for fire, or popping the profiler.

Rubydesic:
lgtm

PR #975, opened 2024-09-29
Re-add Computer Craft compatibility

xiewuzhiying:
Basically consistent with 1.18.2;
`MixinWirelessModemPeripheral` makes the range increase linearly of wireless modems based on world position rather than shipyard position.


Rubydesic:
merged manually

Issue #709, opened 2024-01-28
auto jump being triggered while walking up stairs & weird swimming movement


pearlcrab312:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods

- [X] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

- minecraft 1.20.1
- forge 47.2.0
- valkyrienskies-120-2.1.3-beta.1
- kotlin-4.10.0-all

1. 
If you walk up stairs with blocks next to the stairs,
then angle your camera slightly into the blocks while walking up the stairs the character will jump while walking up the stairs.

vanilla stair climbing: https://imgur.com/a/sg99SlV
valkyrien stair climbing: https://imgur.com/a/I8yTwVN

similar issue found here: https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/530

2.
if you place water on a flat area the water will completely stop you from walking before you reach the water source block. This is different than vanilla water movement.

vanilla: https://imgur.com/a/LBJpSqW
valkyrien: https://imgur.com/a/O5eKcMc

### Issue reproduction

1.
Walk up stairs while facing blocks next to the stairs.

2.
place a water source block in a flat area, then try to walk towards the source block.

Pinkgeneral:
I just noticed it myself! I had no idea this mod mightve been the cause

kragnoth:
have you found a fix for it,  wanted to try out the create mod that let's you have multiple layers of spinning, but it's launching me in the air with stepup

TwentySeventhLetter:
Can also confirm I've encountered this issue, I thought I had left Autojump on somehow, and had no idea it was this mod causing it.

Ryanicus2002:
Can confirm the issue as well, and it also exists in the 1.19.2 version.

gedeknepper:
This also happens when on a horse (riding up a 1-block incline with a wall to the side).
Also occasionally when leading animals up a slope -- they get launched upwards and take small amounts of fall damage.

Serial-Designation-V:
same is happening on 1.20.1, walking up the Stairs and suddendly jumping like Autojump is on

Andy608:
This is also happening for me. Is there a fix for this?

blockninja124:
I believe stairs have been fixed, but water hasn't.

PR #977, opened 2024-10-02
fix water pushing force is too large #709

MoePus:
but should we combine this feature into common instead of implement it separately?


Rubydesic:
merged manually

PR #1006, opened 2024-11-17
Fix for some entities crashing when getting added to a constructed ship.

Andy608:
This change adds a check against the entity's renderer to see if it is eligible for rendering before it calls the render function. This prevents crashes from occurring where an entity may not be ready to render yet, i.e. in the case of Mekanism's Universal Cables.

Also changed the way the java version is set for the project, so developers that have multiple java versions installed can still use this project without needing to reset their java home to java 17.


Andy608:
See Mekanism thread for vis: https://github.com/mekanism/Mekanism/pull/8260

Rubydesic:
Please leave the Java release/compiler options in place. The toolchain stuff is fine but you don't need to remove the previous settings.

Rubydesic:
I don't really understand why this patch does anything. We're calling `blockEntityRenderDispatcher#render` (which is what vanilla does), which already checks #shouldRender. This is actually necessary for Mekanism not to crash?

Andy608:
I did more testing and found that this causes beds and other uniquely rendered things to be invisible so this isn't a good fix.

I think my original null check in Mekanism is the best approach to solve this.

Rubydesic:
Closing for now, then.

PR #1010, opened 2024-11-21
1.19.4/main

millennIumAMbiguity:
create doesn't have a public 1.19.4 among other mods.

not everything is done, but I'm sharing my progress


millennIumAMbiguity:
Current roadblock

```
Valkyrien-Skies-2\common\src\main\java\org\valkyrienskies\mod\mixin\server\MixinMinecraftServer.java:352: error: incompatible types: EntityDimensions cannot be converted to Entity
            return PortalShape.createPortalInfo(destLevel, foundRectangle, axis, vec3, entityDimensions, deltaMovement, 0.0f, 0.0f);
```

millennIumAMbiguity:
Currently compiles on both Fabric and Forge, but there are still issues regarding Mixin

millennIumAMbiguity:
Should be merged from 1.20 instead of 1.19. Will not finish this pr.

PR #1021, opened 2024-12-13
Use seated pos when checking entity distance

Endalion:
**CHANGE, REASON**
- Use the seat position where valid in entity distance checks 
   -  Utilise the seat position in ship for distance checks which is guaranteed to be accurate _for checks on the same ship_
- `Entity::getEyePosition` without partial tick uses VS2 modification as well
   -  pick does not use entity distance checks instead it uses `IForgePlayer::canReach`, which uses `Entity::getEyePosition()`, which is not mixin'ed by VS2 (only `Entity::getEyePosition(F)` is)

**KNOWN EFFECTS**
- Interacting with a moving ship while seated on it is significantly more reliable, up to 1km/s (tested)

**CAVEATS**
### **NOT TESTED ON FABRIC**
- ~~no guarantees when player is not seated~~ See: #1027
- not tested to work between ships, I have no clue when that'd ever apply in normal gameplay

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/entity/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/entity/MixinEntity.java
@@ -125,6 +125,29 @@ public abstract class MixinEntity implements IEntityDraggingInformationProvider 
         cir.setReturnValue(newEyePos);
     }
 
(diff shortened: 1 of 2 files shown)


ThePlasticPotato:
Merged into #1027 as of commit [b84d58a](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1027/commits/b84d58ad5fcb6a76c6119826b74ae88214c772fe).

PR #1039, opened 2024-12-27
Tiny Cleanup of FTB Chunks Mixin

TechTastic:
Just a tiny change to use `VectorConversionMCKt#toJOMLD` and rearrange some bits


TechTastic:
Updated to use new API

TechTastic:
Updated to use FTB's repo to get only FTB Chunks rather than using CurseMaven

PR #1074, opened 2025-01-21
Moving TIS-3D Mixins to Common and Updated Dependency

TechTastic:
Updated TIS-3D Dependency and moved it to Common as well as moved TIS-3D Mixins and extra bits from Forge to Common


TechTastic:
huh, thats a strange failed test, it does work on Fabric...

PR #1077, opened 2025-01-22
[1.20.1] Modular Force Field System Compat

TechTastic:
This PR adds a new Mixin (`MixinClientPacketHandler`) that fixes the Beam Particle's target position to point ot the proper Worldspace position rather than a Shipyard position

There is also a README included and a new `modCompileOnly` dependency for MFFS


TechTastic (commented):
Rebased to `1.20.1/2.3`

PR #1079, opened 2025-01-22
[1.20.1] Fixing Epic Fight crash

TechTastic:
I added a new `FracturedBlockStateInfoProvider` for handling Epic Fight's custom `FracturedBlockState` which is only used on the client and is the reason for the crash


TechTastic:
Closing to open another on the proper branch

PR #1083, opened 2025-01-25
fix 1.20.1 not compiling due to cccompatibility

cylian914:
error was: ```error: Unable to locate obfuscation mapping for @WrapOperation target getRange```


blockninja124:
I haven't tested this specifically, but the change looks good to me. 

(I doubt the change could break anything, and it probably _is_ needed since we don't want to remap for CC mixins)

blockninja124 (dismissed):
I haven't tested this specifically, but it looks good to me.

(I doubt the change could break anything, and it probably is needed since we don't want to remap for CC mixins)

cylian914:
i completely forgot i made this is a while ago
its already fixed iirc so im closing the pr

PR #1159, opened 2025-05-06
Fix null position when exiting seats

Endalion:


PriestOfFerns:
a pr description would be nice

PR #1168, opened 2025-05-10
[1.20.1] FTB Chunks Fix

TechTastic:
I updated the FTB Chunks dependency to use current latest and changed to using FTB's maven instead of CurseMaven

I also finally fixed `MixinClaimedChunkManagerImpl` so it no longer crashes


TechTastic (commented):
Rebased on current `1.20.1/2.3` and tested on Forge and Fabric

blockninja124:
Is this also the MFFS compat PR? The commits seem to be mixed in

thetheaplant (approved):

blockninja124 (approved):

PR #1169, opened 2025-05-10
[1.20.1 Forge] MFFS Compat

TechTastic:
Modular Force Field System Compatibility Mixins

- `MixinClientPacketHandler`
  - Fixes the target position of the Beam Particle if it is on a Ship pointing it to the correct Worldspace position
- `MixinFrequencyGrid`
  - Changes a distance check within `FrequencyGrid` so that Ship-bound Forton devices are accounted for within the set range of other devices both in-world and on other Ships
  
Other Compatibility Changes
- Added Masses and physics values to MFFS blocks


TechTastic:
Ok, so there is more to do here

TechTastic:
Got it all working including the new `MixinFrequencyGrid` thanks to `The...Why...Even...How?` in the NeoForge discord!

TechTastic (commented):
mffs-compat-test

TechTastic:
https://streamable.com/14enh8

PR #1170, opened 2025-05-10
[1.20.1] Moving TIS-3D Compat to Common

TechTastic:
Updated TIS-3D Dependency and moved it to Common as well as moved TIS-3D Mixins and extra bits from Forge to Common


TechTastic (commented):
Tested and works on both Forge and Fabric!

tis-3d-test

blockninja124 (approved):
LGTM. Did a little testing and it all seemed to work forge+fabric (and if TIS isn't installed). I'm not knowledgeable enough on TIS to properly test the infrared module, but I'm sure techtastic did that already

thetheaplant:
sure does

PR #1171, opened 2025-05-10
[1.20.1 Forge] Epic Fight Fix

TechTastic:
EpicFight Compat Class

- FracturedBlockStateInfoProvider
    - Epic Fight adds its own custom BlockState called FracturedBlockState that only exists on the Client and cannot be handled by MassDatapackResolver


TechTastic (commented):
https://streamable.com/lv4p78

I was finally able to test my fix!

PR #1172, opened 2025-05-11
[2.3] Weather2 compat

PriestOfFerns:
Adds alex's weather 2 compat to the 2.3 branch. Code is more or less 1:1 copied from 1.20.1/main


PriestOfFerns:
It appears to work when doing out-of-dev testing

TechTastic (changes requested):
It seems Ships share forces somewhat under Weather2 compat
https://streamable.com/6duxf9

PriestOfFerns:
> It seems Ships share forces somewhat under Weather2 compat https://streamable.com/6duxf9

[Should be fixed now](https://streamable.com/ffehdx)

PR #1185, opened 2025-05-27
Fix injection point for new shipyard entitiy

Endalion:
# Issue
* Contraption wings do not work after world reload unless re-assembled.

This is due to the injection point for new shipyard entity only catching entities **added to the level**, so entities loaded from chunk savedata do not get intercepted. 
This results in the wing initializing code not being run.

# Solution
Move the injection point into `PersistentEntitySectionManager:addEntity`, where all newly loaded entities can be intercepted.

Unknown how this will effect other shipyard entities logic, but so far no issues in regular gameplay usage.


Endalion:
I have attempted but failed to wrap the entity check in the original method being injected. Unsure if this will cause issues down the line. Otherwise insofar the current iteration is stable in regular usage.

PR #1198, opened 2025-06-15
Create 6 support backported from 1.21.1/main (both fabric and forge this time)

m3t4f1v3:
compiles and mostly works in both forge and fabric, issues with contraptions rendering on fabric specifically

pretty much https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1190 but better


Justadude35:
> It compiled the jar, but when running mc it says that it's not a valid mod file. Log says: net.minecraftforge.fml.loading.moddiscovery.InvalidModFileException: Illegal version number specified unspecified (valkyrienskies-120.jar)

I forgot I was tinkering with the gradle.build file yesterday and deleted the  "version = mod_version + "+" + gitRevision.substring(0, 10)" because it caused an error -_-'

Adding it back fails the build, through gradlew build and gradlew.bat with a:
FAILURE: Build failed with an exception.

* Where:
Build file 'build.gradle' line: 26

* What went wrong:
A problem occurred evaluating root project 'valkyrienskies'.
> begin 0, end 10, length 0

Justadude35:
> which folder were you copying in from

forge/build/libs

blockninja124:
For future reference, the mod version property fails to compile because `gitRevision` doesn't always exist on the local copy. You can simply change your local copy from
`version = mod_version + "+" + gitRevision.substring(0, 10)`
to
`version = mod_version`

m3t4f1v3:
turns out the issues with contraptions rendering on fabric is a create fabric issue by itself (its unreleased so can't blame them)

prufaraSS:
trackworks crashes [mclogs-reason](https://mclo.gs/kitqfLL) [mclogs-full](https://mclo.gs/O5YXzM7)

thetheaplant:
@prufaraSS trackwork does not support create 6, please do not comment unrelated issues.

ThePlasticPotato:
Combined into #1212

PR #1200, opened 2025-06-16
Fix forge 47.4.2 

jcm236:
Fix not being able to interact with ships

### DIFF
--- a/forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/forge_interact/MixinIForgePlayer.java
+++ b/forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/forge_interact/MixinIForgePlayer.java
@@ -35,7 +35,7 @@ public interface MixinIForgePlayer {
                 pos.getY() + 0.5,
                 pos.getZ() + 0.5,
                 eyes.x, eyes.y, eyes.z
-            ) < reach * reach;
+            ) <= reach * reach;
         } else {
             return true;
         }
@@ -54,7 +54,7 @@ public interface MixinIForgePlayer {
                 pos.getY() + 0.5,
                 pos.getZ() + 0.5,
                 eyes.x, eyes.y, eyes.z
-            ) < reach * reach;
+            ) <= reach * reach;
         } else {
             return true;
         }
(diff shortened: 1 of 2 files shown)


blockninja124:
It was missing an import but I tested and it works! Full ship functionality on forge 47.4.2

forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/forge_interact/MixinIForgePlayer.java
```diff
@@ -39,6 +40,26 @@ default boolean canReach(final BlockPos pos, final double padding) {
             return true;
         }
     }
+    
+    /**
+     * Include ships in server-side distance check when player interacts with a block.
+     */
+    @Overwrite(remap = false)
```
zyxkad, line 47: Please Inject not Overwrite, it will broke a lot stuffs

zyxkad (changes requested):

forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/forge_interact/MixinIForgePlayer.java
zyxkad: Never mind, just realized you can't use injector for interface

forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/forge_interact/MixinIForgePlayer.java
```diff
@@ -39,6 +40,26 @@ default boolean canReach(final BlockPos pos, final double padding) {
...
+    default boolean canReachRaw(final BlockPos pos, final double padding) {
+        if (VSGameConfig.SERVER.getEnableInteractDistanceChecks()) {
+            double reach = self().getAttributeValue(ForgeMod.BLOCK_REACH.get()) + padding;
+            final Vec3 eyes = this.self().getEyePosition();
+            return VSGameUtilsKt.squaredDistanceBetweenInclShips(this.self().level(),
+                pos.getX() + 0.5,
+                pos.getY() + 0.5,
+                pos.getZ() + 0.5,
+                eyes.x, eyes.y, eyes.z
+            ) < reach * reach;
```
zyxkad:
```suggestion
            ) <= reach * reach;
```

Forge somehow changed `<` to `<=`, same for above `canReach` but I cannot comment there

zyxkad (changes requested):

PR #1203, opened 2025-06-17
Updated MixinSoundPhysics

GuiltyRegicide:
Modified code to use processSound instead of onPlaySound to reflect whats used in the more recent version of SoundPhysics


blockninja124:
Tested and this PR doesn't work yet unfortunately. I believe the `processSound` isn't even being called on the clients for the shipyard sounds, so we probably need another mixin earlier on. I'll convert this PR to a draft for now, mostly so it doesn't accidentally get merged, and you can make it ready for review once you think its working.

This compat will be great to have fixed, glad someone is working on it!

PR #1222, opened 2025-07-03
Fix vs commands

blockninja124:
`VSCommandSource` is no longer, we're now using the base Minecraft `SharedSuggestionProvider` with some extension functions to handle the impls which the command system seems to be much happier about.

Other changes:
- I removed the ability to use @v in the normal /tp command, since it broke the vanilla command quite a bit. If someone can find a way to do it without being so janky, then we can add it back.
- `/vs ship rename` is now just `/vs rename`, and all of the functionality in `/vs ship delete` was already in `/vs delete` anyway.
- `/vs get-ship` will now inform you on the ship id as well as slug, so using `id` in a ship selector is actually viable now. Although the chinese and korean lang files will need updating to show the id, until then they will still only show the slug.
- The vector argument type previews `(0 0 0)` before you type `(` so its easier to know what it wants


blockninja124:
I think I accidentally pinged some random guy named V, sorry about that lol!

thetheaplant (approved):

ThePlasticPotato (approved):
fire

PR #1226, opened 2025-07-05
Ship/world-aware hose pulleys (compatible both with create v0.5 and v6), enchanting table tweaks and a minor Sodium fix

zaafonin:
Hose pulleys can pull/push liquids from ships to world, a feature known from VS Addition. It is however not exactly reliable and not yet compatible with Create v6. My variant might be a bit more computationally expensive but it seems very reliable.

Enchanting tables placed on ships now orient their floating book towards the player correctly. They are also aware of world-positioned bookshelves. Bookshelves of other ships are not checked but this should be an easy addition.

Previously when using Sodium block animation was broken on ships, as ship chunks were not in render lists used for updating sprites. This fix basically copies what is already done for rendering block entities which also suffered from ships not being included in these render lists. Maybe there is a deeper issue and fixing it would solve both problems so that both mixins will be unnecessary.


ThePlasticPotato (dismissed):
Looks good to me!

blockninja124:
Does this prevent hose pulleys from working ship -> same ship? I remember even with VS additions modes on the hose pulley, a ship to itself was quite unreliable.

zaafonin:
It should not, if a hose hits a "bottom" on the same ship it uses that block and doesn't check against the world. I'll test that again soon though

zaafonin:
> Does this prevent hose pulleys from working ship -> same ship? I remember even with VS additions modes on the hose pulley, a ship to itself was quite unreliable.

Tested it, works fine. I wonder if this feature should be configurable, after all vs json config already has config entries for mod integration

TechTastic (dismissed):
Works for me on Fabric with both normal and scaled Ships, @getItemFromBlock is testing Forge right now

getItemFromBlock:
For the reference, we talked about some bugs and improvements over on discord. I will wait for @zaafonin 's changes before my final review. Other than that looks good to me 👍

zaafonin:
X-to-ship pumping should be more reliable now. I couldn't replicate the wrong orientation of sculk particles but I found shifting position +0.5 was not necessary so now their destination position is correct. The mixin for adjusting ench table book orientation for rotated ships was reimplemented, modern technology is surprisingly good at it :( The feature with ship ench tables recognizing world bookshelves was removed, I think this is excessive especially since the bookshelf check happens every time the table tries to spawn particles, which is rather often.

ThePlasticPotato:
has been sitting for 2 months unmerged but seems fine, also i approved it in the past

PR #1228, opened 2025-07-05
Fix them balls

blockninja124:
The debug physics entity ball renderer no longer causes crazy FPS lag, since it uses a proper entity model and has a lot less faces.

Before

https://github.com/user-attachments/assets/8d3bc5f4-553d-484e-94ab-403daadec03b

After

https://github.com/user-attachments/assets/962dc121-36cd-45b9-8f98-cc2390cc4e64

While this PR is done, it probably shouldn't be merged until `1.20.1/main` is a bit more fixed


common/src/main/kotlin/org/valkyrienskies/mod/client/VSPhysicsEntityRenderer.kt
```diff
@@ -2,72 +2,46 @@ package org.valkyrienskies.mod.client
...
-                fallingBlockEntity.blockPosition()
-            ) || blockState.renderShape == INVISIBLE
-        ) {
-            return
-        }
 
-        val renderTransform = fallingBlockEntity.getRenderTransform(
-            ((Minecraft.getInstance() as IShipObjectWorldClientProvider).shipObjectWorld as ClientShipWorldCore)
+        val renderTransform = mob.getRenderTransform(
+            ((Minecraft.getInstance() as IShipObjectWorldClientProvider).shipObjectWorld as ShipObjectClientWorld)
```
TechTastic: `ShipObjectClientWorld` no longer exists according to IntelliJ

common/src/main/resources/assets/valkyrienskies/models/debug_physics_entity.bbmodel
TechTastic, line 1: Why `bbmodel` and not an entity JSON model?

TechTastic (changes requested):
A small change is needed but it looks fine otherwise

common/src/main/resources/assets/valkyrienskies/models/debug_physics_entity.bbmodel
blockninja124: Because the model is exported to a Java class, I just included the bbmodel so that others can tweak the model later. (The java class can't be imported as a model after being exported)

common/src/main/kotlin/org/valkyrienskies/mod/client/VSPhysicsEntityRenderer.kt
blockninja124: Ah something must have changed

common/src/main/resources/assets/valkyrienskies/models/debug_physics_entity.bbmodel
TechTastic: Fair enough

common/src/main/kotlin/org/valkyrienskies/mod/client/VSPhysicsEntityRenderer.kt
TechTastic: image

TechTastic (dismissed):
LGTM

ThePlasticPotato:
your balls have passed the vibe check

PR #1256, opened 2025-07-26
[1.20.1/2.3] Modern lang file from 1.20.1/main

LogicWheat:
Same as #1236


blockninja124 (approved):
Looks good to me, man the lang file was such a mess before

PR #1257, opened 2025-07-26
Compat for Connectible Chains

zaafonin:
Simple compatibility for [Fabric](https://modrinth.com/mod/connectiblechains) and [Forge](https://www.curseforge.com/minecraft/mc-mods/connectible-chains-forge) versions of Connectible Chains, a mod allowing fences to be connected with chains and leads.

Pictured: world-to-world connections still functional, world-to-ship and ship-to-ship connections also rendered correctly.

"Chain knots" (entities similar to leash knots of vanilla) are now shipyard entities which allows them to occupy shipyard coordinates and correctly check whether they're still attached to the fence. This was not the case before this PR (attaching a chain to a ship fence resulted in it being broken). Coordinate recalculation is necessary to avoid a crash from attempting to render a 20 million block long chain.

The bug with chains not rendering if the knot is out of the view is not fixed, it faithfully matches the same bug that occurs with vanilla leads on ships. To keep my PR small I've chosen to fix it later but if really necessary I can try to fix it now.

While the Forge port is seemingly licensed GPLv3, the [specific file from Forge version](https://github.com/lilypuree/ConnectibleChains/blob/1.20.1/src/main/java/com/lilypuree/connectiblechains/client/render/entity/ChainKnotEntityRenderer.java) I mixin into is licensed as LGPLv3 as confirmed by the dev of this fork. The mixin does not copy functionality implemented in the mod and only handles position transformations.


blockninja124:
LGPL on that file we mixin means that we are allowed to do this, right?

zaafonin:
> LGPL on that file we mixin means that we are allowed to do this, right?

Verbal (DM) permission to distribute a mixin into this file was confirmed from the dev of Forge version of the mod. Whether a mixin is derivative code or simply dependent on the library is an interesting question but I suspect it falls under derivative code. IMO we should contact the dev of the original Fabric version to clear up the situation

zaafonin:
Fixed chains (and leashes) being erroneously culled when one knot is in the world but the other is on a ship. Also fixed name tags of shipyard entities like armor stands or chain knots not rendering.

ThePlasticPotato:
build failure, will merge when that is fixed

zaafonin:
Maven mirror is missing `curse.maven:connectiblechains-415681:7148381`, not something I handle

zaafonin:
Updated this PR with a couple of fixes for scaled ships, also chains now obey world gravity, new CompatUtil methods are used. Fixed some remap issues so this thing works on Fabric dev, Forge dev, Fabric dev server, Forge dev server, Fabric prod and Forge prod. I consider this PR finished on my side

PR #1260, opened 2025-07-30
fix CC compat issue in release envionment

zyxkad:


blockninja124 (approved):
Looks good to me, although I don't remember exactly what issue this fixes. Turtles throw a java error when moving or something?

zyxkad:
> Looks good to me, although I don't remember exactly what issue this fixes. Turtles throw a java error when moving or something?

yes, it's because the mixin is not remapped

TechTastic (approved):
LGTM

SquidDev:
Would it be possible to get a new release out with this fix? I've had several crash reports due to this bug.

thetheaplant:
@SquidDev We will probably be able to get a release out in the next few days, apologies for the erroneous issues.

C0RR0DE:
yes, having this done would help my modpack stay at having MOST of the most in it activated (lots are deactivated due to lag, and VS2 due to incompat).

thetheaplant:
@SquidDev This fix has been released in beta.10

Issue #1252, opened 2025-07-24
Compatibility issue between Valkyrien Skies and Mekanism - Crash when opening Dynamic Tank and Reactor interfaces


hjl514:
### Mod Name

Valkyrien Skies and Mekanism

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [x] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

When both Valkyrien Skies and Mekanism are installed, trying to open the Dynamic Tank or Reactor interfaces causes the game to crash immediately.

### Issue reproduction

Install Valkyrie Sky, mekanism and Mekanism Generators mod
Start a new game and create a world.
Place a Mekanism Dynamic Tank or Reactor.
Right-click on the Dynamic Tank or Reactor to open its interface.
The game crashes.

### Logs

[crash-2025-07-24_09.48.23-client.txt](https://github.com/user-attachments/files/21398466/crash-2025-07-24_09.48.23-client.txt)

blockninja124:
Have you tested to see if [Kontraption](https://modrinth.com/mod/kontraption) fixes this issue? I believe it fixes a few mekanism crashes with VS, but I don’t know which ones specifically.

hjl514:
> 您是否测试过 [Kontraption](https://modrinth.com/mod/kontraption) 是否解决了这个问题？我相信它修复了 VS 的一些机制崩溃，但我不知道具体是哪些。

Hi, thank you for your suggestion. I have tried using Kontraption version 0.0.4 along with Valkyrien Skies and Mekanism, but the game still crashes when I attempt to open the dynamic tank and reactor interfaces. I'm currently using Mekanism version 10.4.16 and Valkyrien Skies version 2.3.0-beta.7.  Could you please help me further investigate this issue?

hjl514:
Hi, thank you for your suggestion. I have tried using Kontraption version 0.0.4 along with Valkyrien Skies and Mekanism, but the game still crashes when I attempt to open the dynamic tank and reactor interfaces. I'm currently using Mekanism version 10.4.16 and Valkyrien Skies version 2.3.0-beta.7. Could you please help me further investigate this issue?

blockninja124:
[...]

PR #1261, opened 2025-07-30
Duct tape ahh fix for #1252

zaafonin:
Thanks to @hjl514 for discovering the bug described in #1252 (dynamic tanks and other multiblocks crash at rendering if they're located on ships). Here is a fix for this bug plus a couple other minor issues. This fix is functional but the bug suggests there is a deeper issue with Mekanism multiblocks on ships, possibly some desync. Once the culprit is found and fixed this fix will not be necessary.

I've updated Mekanism and added it to runtime. Tell me if you'd rather keep it compileOnly, but it doesn't work with dev env if put as a jar so for developing further compat I think it's best to have it right here.


hjl514:
Thank you for your response and the fix you've suggested. I will try updating Mekanism and adding it to the runtime. Regarding whether to keep it as compileOnly, I would prefer to follow your professional advice. If placing it as a jar in the development environment is more conducive to compatibility and future development, then I will act according to your recommendation. Thank you again for your help, and I look forward to further guidance.

ThePlasticPotato:
until the core issue is resolved im unsure if i can justify merging this

forge/build.gradle
```diff
@@ -138,7 +138,7 @@ dependencies {
     modCompileOnly("curse.maven:common-capabilities-247007:4987207")
 
     // Mekanism
-    modCompileOnly ("curse.maven:mekanism-268560:4644795")
+    modImplementation ("maven.modrinth:mekanism:${mekanism_version}")
```
Rubydesic: Would rather keep this modCompileOnly. Frankly would like to do this to other mods that are included by default but lets not make the bloated dev environment worse than it already is

LogicWheat:
@zaafonin  can you resolve the conflicts? thanks!
this bug broke 3 worlds of mine >:(

Issue #1161, opened 2025-05-07
Not working with Encased fan processing


catoxide:
### Mod Name

Create/Create connected

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [x] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

On Valkyrien Skies ships, Create Encased fan processing does not work with blaze burner in front nor catalyst from Create Connected which should be working as typical fluid works.

### Issue reproduction

as described

catoxide:
*ps: it just seems not working... seems no wind passing from it like normal blocks... idk whats wrong with it...

blockninja124:
is it in general, or only on ships? (assembled physics objects)

catoxide:
just on ships only
i'm sure that's maybe VS's falut as blaze burner also doesn't work
already issued them
(sorry for not replying timely as login takes efforts

catoxide:
oh I didn't reach to github for weeks for net problems
sorry for not reply timely
i'll show you the case i just record on my server
i'll stick my link below

catoxide:
【For Valkyrien Skies bug report use】 https://www.bilibili.com/video/BV1E7EYzREdE/?share_source=copy_web&vd_source=f390ab5b938f90e1dfcce3f516002af0 
sry no english(
the recipe is vanilla and not changed

blockninja124:
Hmm I wonder if the blaze burner only works for encased fans if Create: Connected is installed. 

If so, this is a compat issue. But if create on its own allows blaze burners to function like that, then its possible Valkyrien skies isn't checking for full/non-fluid blocks for fans?

catoxide:
don't really get what is the first sentence meaning?
these blocks should be "wind-passable" for fans? I think yes?
it is just a minor use of blaze burner i think?
i don't know how create make this though?
but i believe something changes about these blocks for "wind" doesn't seems to go through?

catoxide:
btw when i am updating vs from beta-5 to beta-7 it brokes
beta-6 works

blockninja124:
> btw when i am updating vs from beta-5 to beta-7 it brokes beta-6 works

If you have VS-addition installed, it needs a special version for VS beta.7

blockninja124:
Should be fixed in beta.11. Please open a new issue if you still experience this after updating.

PR #1263, opened 2025-07-31
Fixes for #1161, #1285 (Create), #1278, #1279

zaafonin:
The reason for #1161 (encased fans on ships failing to bulk process recipes that are triggered by putting a full block in front of a fan, like a blaze burner or fan catalysts from addons): VS2 raycasts to check if a fan is obstructed by any ship. This raycast does not exclude fan's own ship, so if 1) a fan is on a ship and 2) there's a full block in front of a fan it, the air flow will stop. Now we first check obstructions in the Create way (same ship), then check for ship obstruction while excluding own ship.

Also bulk processing world items by ship fans was broken in 2.5, probably because of misported mixins. Now these mixins are version-agnostic so fans work again both on v0.5.1 and v6.

For the second commit: afaik @ThePlasticPotato said it's ok to uncomment these lines and it was already done in some other branch. Correct me if I'm wrong.


TechTastic (changes requested):
While this does fix Fans working on Ships, it seems the processing abilities do not affect items in worldspace from a Ship
image

It also seems to not work on scaled Ships
image

It also seems that Ships block fans too early
image

*Tbh, these last two issues are likely the same issue*

TechTastic:
image

Ok, so it *does* work on scaled Ships but without a block close to it and a LOT closer to the fans

zaafonin:
Should be fixed now. Fan strength and distance is scaled matching the ship scale, but the distance is consistent so bulk processing works. There might be rare problems with processing worldspace items on miniships as the air current AABB is smaller and there's a chance for an item to miss it completely (at least that's how I understand it), but most of the time it's fine. Depots work as well

zaafonin:
https://github.com/user-attachments/assets/bea39d6a-9fbd-44e4-87bc-8127d071ddd9

Now everything should work on tilted ships, including particles.

zaafonin:
Fix for #1285, also version-independent.

https://github.com/user-attachments/assets/a90fae84-2096-4e0b-8536-e5cda44c1f4a

zaafonin:
Summary for this PR as asked:
- #1161: fixed, parts of MixinAirCurrent were rewritten so obstructions are handled correctly. Also fixed processing on scaled ships
- #1285: fixed, the culprit was Entity#stepOn not firing when entities stepped on ships. Coincidentally this fixes #1279 (sneaking on ships gives extra speed) and probably other bugs arising from players not technically stepping on ship blocks.
- #1278: fixed, now placement checks work for worldspace players and shipspace blocks. This uses an OBB intersection check (worldspace player AABB against shipspace block AABB with ship-to-world transformation) because a "regular" AABB transform would inflate it, yielding a lot of false positives. However, player's AABB is slightly deflated to allow some wiggle room: clientside and serverside ship transforms can be noticeably different, so a precise check can succeed on one side but fail on the other. An occasional false negative is bad but frequent ghost blocks are worse.
- Shipspace funnels barely ever accepting dropped items: fixed, now uses a simpler injection point, works on both Create versions.
- Entities not interacting with shipyard fluids on Fabric: fixed, the functional mixin was there but not enabled in config. Maybe it wasn't enabled for some obscure reason?

ThePlasticPotato:
lgtm

PR #1264, opened 2025-08-01
Compatibility fixes for some mods

CodeCrafterJerry:
It can run normally, solving the problem of creating large cannons using embedded, sodium, chloride, and development versions. It also fixes the issue of Iris and VS incompatibility causing startup crashes, as well as the problem of wings not being able to rely on physical mechanical power to work. In addition, it also solves the problem of game crashes caused by using flywheels to render dynamic structures in physical structures


blockninja124:
I don't think it's worth making pull requests for 1.21 at this time, since the 1.21 branch was made somewhat incorrectly (based on a weird version of VS 2.3) and will probably have to be redone from scratch once 2.5 is complete. That being said, I don't know other contributors thoughts on this, so this could still get merged.

PR #1266, opened 2025-08-03
Change vscore version and some common fixes

CodeCrafterJerry:
Changed the VSCORE version and synchronized some code from 1.20.1 common,It can run normally, solving the problem of creating large cannons using embedded, sodium, chloride, and development versions. It also fixes the issue of Iris and VS incompatibility causing startup crashes, as well as the problem of wings not being able to rely on physical mechanical power to work. In addition, it also solves the problem of game crashes caused by using flywheels to render dynamic structures in physical structures


LogicWheat:
dynamic structures → Contraption (Create Mod)
physical structures → Ship

PR #1268, opened 2025-08-03
1.21.1/main

CodeCrafterJerry:
Changed the VSCORE version and synchronized some code from 1.20.1 common,It can run normally, solving the problem of creating large cannons using embedded, sodium, chloride, and development versions. It also fixes the issue of Iris and VS incompatibility causing startup crashes, as well as the problem of wings not being able to rely on physical mechanical power to work. In addition, it also solves the problem of game crashes caused by using flywheels to render Contraption (Create Mod) in Ship


build.gradle
```diff
@@ -91,6 +91,14 @@ subprojects {
         mavenCentral()
         if (!project.block_external_repositories) {
             mavenLocal()
+            maven {
```
PriestOfFern, line 94: Why are you adding this supposed ForgifiedFabricAPI maven if "https://maven.su5ed.dev/releases" is the maven listed on the github? You also already have "https://maven.su5ed.dev/releases" added lower down

build.gradle
```diff
@@ -116,10 +124,12 @@ subprojects {
             maven { url = "https://maven.tterrag.com/" } // Registrate, Forge Create and Flywheel
             maven { url = "https://maven.cafeteria.dev/releases" } // Fake Player API
             maven { url = "https://maven.jamieswhiteshirt.com/libs-release" } // Reach Entity Attributes
+            maven { url = "https://maven.ftb.dev/releases" } // FTB Mods
             maven {
                 name = 'NeoForged'
                 url = 'https://maven.neoforged.net/releases'
             }
+            maven {url = "https://maven.su5ed.dev/releases" }
```
PriestOfFern, line 132: Why are you adding a forgified fabric api maven, if you're not using forgified fabric anywhere?

forge/libs/createbigcannons-5.9.1-dev.jar
PriestOfFern, line 1: why are you adding cbc as a jar, instead of adding it as a gradle dependency?

forge/build.gradle
```diff
@@ -59,16 +59,22 @@ dependencies {
         transitive = false
     }
     //modCompileOnly("curse.maven:rubidium-574856:4024781")
-    modCompileOnly("maven.modrinth:embeddium:${embeddium_version}")
+    modCompileOnly("maven.modrinth:sodium:mc1.21.1-0.6.13-neoforge")
+    modCompileOnly("maven.modrinth:chloride:NEOFORGE-mc1.21.1-v1.7.2")
+    modImplementation("maven.modrinth:embeddium:1.0.15+mc1.21.1")
```
PriestOfFern, line 64: you shouldn't be adding hardcoded versions directly to the gradle. There should a gradle property for each mod.

PriestOfFern (changes requested):
This is an extremely cursed PR

PriestOfFern:
Also, I assume most of the changes come from you merging 1.20.1/main into the 1.21 branch. If so, you must have merged in an extremely weird way, leading to the commit history being clobbered

TechTastic (commented):
No wonder Flywheel doesnt crash, you removed every single Flywheel compatibility Mixin

Did you test any Create-specific stuff on Ships? If not, you might want to as these were important Mixins for making Flywheel stuff render correctly on Ships

forge/build.gradle
CodeCrafterJerry: This is a terrible code, I don't understand Gradle

CodeCrafterJerry:
I apologize for submitting this PR prematurely

PR #1303, opened 2025-09-15
fix newly create ship will not render untill player rotated its head

zyxkad:
It may have performance impact but not noticeable.


ThePlasticPotato:
lgtm

PR #1314, opened 2025-09-30
Flywheel Visual & Create GhostBlockEntity compatibility

chj990707:
Rendering of create GhostBlockRenderer on vs ship is implemented.
- GhostBlockRenderer checks for BE position if it is on a ship and transforms itself.
- A bit rubber-bands when the ship is moving. Suspected if I called (tick) Transformation instead of RenderTransformation but I didn't. Might need polishing.

Rendering of flywheel BlockEntityVisuals on vs ship is implemented.
- Originally BlockEntityStorage class of flywheel holds BlockEntityVisuals for rendering, and they apply correct transformation to their Instances. 
- To implement rendering on ship, BlockEntityStorage is injected to check if the newly added BE is on a ship and register the relation.
- BlockEntityStorage also creates & updates VisualEmbedding, subclass of VisualizationContext which is in the original flywheel API to support arbitrary transformation of multiple visuals, thereby creating a 'meta-visual'.
- BlockEntityStorage.vs$updateAllShips() should be called to update VisualEmbedding transformations, but Storage classes themselves don't have tick() or render() methods.
- Instead, it is called by VisualizationManagerImpl on every render() calls, and updates the transformation of each BlockEntityVisuals on ships.


chj990707:
NOTE: I intentionally left other file modification from commit history to prevent possible merge conflicts. So this PR itself doesn't assure build success. All other changes were superficial, only refactoring changed package directories and temporarily deleting non-functional injectors.

chj990707:
Will resolve merge conflict and submit again.

PR #1315, opened 2025-10-01
1.20.1/2.3 create6

chj990707:
Rendering of create GhostBlockRenderer on vs ship is implemented.

- GhostBlockRenderer checks for BE position if it is on a ship and transforms itself.
- A bit rubber-bands when the ship is moving. Suspected if I called (tick) Transformation instead of RenderTransformation but I didn't. Might need polishing.

Rendering of flywheel BlockEntityVisuals on vs ship is implemented.

- Originally BlockEntityStorage class of flywheel holds BlockEntityVisuals for rendering, and they apply correct transformation to their Instances.
- To implement rendering on ship, BlockEntityStorage is injected to check if the newly added BE is on a ship and register the relation.
- BlockEntityStorage also creates & updates VisualEmbedding, subclass of VisualizationContext which is in the original flywheel API to support arbitrary transformation of multiple visuals, thereby creating a 'meta-visual'.
- BlockEntityStorage.vs$updateAllShips() should be called to update VisualEmbedding transformations, but Storage classes themselves don't have tick() or render() methods.
- Instead, it is called by VisualizationManagerImpl on every render() calls, and updates the transformation of each BlockEntityVisuals on ships.
- Changes made in f2b39f12 was reverted to remove the duplicate implementation, and confirmed the functionality.

NOTE: This PR intentionally excluded changes made in other files, which were just package renamings, disabling injections missing targets, and dependency version change. It won't ensure the build success.


blockninja124:
PRs changing the dependency number, fixing packages, and other changes are totally fine btw. The only thing I want to avoid (although it might still happen, from myself or PRs) is disabling stuff that isn't fixed purely to get a compilation.

But with how it is right now, even those changes might be better than nothing 🤔

blockninja124 (approved):

chj990707:
- Relying on VisualizationManagerImpl.render to update the ship transform to the embedding is moved to FlwEvents, removing potential danger of mixin to impl class.

chj990707:
- Mixins to flywheel backend classes are removed. It now has 2 Mixins to API classes, each for decorating visualizer and Entity visibility check.
- Works with vanillin rendering.
- Visual Embedding operation is synchronized.

PR #1323, opened 2025-10-17
Arrow and relativity

chj990707:
Adds relative movement to the dragged entities.
- Note: used previously existing variables in the dragging information.
Projectiles(e.g. arrows) can be stuck on ship by moving to the shipyard.
- Once stopped colliding with the ship or picked up they will leave the shipyard too.


chj990707:
I will seperate the relative velocity and shipyard projectiles as 2 branches for the ease of merging into other branches and tracking.

PR #1324, opened 2025-10-21
Sodium render section distance sorted

chj990707:
Sorts render sections(chunks) being rebuilt on ship with corrected distance. Formerly the ships were always added last, causing them to always render last.
This includes sorting non-ship chunks too. While the sodium field name implies the rebuild list is sorted, the exact side effect is unknown.


chj990707:
Sorry, submitted with wrong branch. will fix and repost.

PR #1325, opened 2025-10-21
Projectiles to shipyard

chj990707:
If any projectile collides with ship block, it'll be sent to shipyard. If said projectile stops colliding with the ship blocks, it'll return to worldspace.
This PR is meant to fix arrows not sticking on ship, but implements it by generic projectile, so other things like fishing hooks also work as the same.
Arrows and tridents can be picked up normally thanks to previous existing code. Loyal tridents return to the owner too, while the initial velocity is a bit off.


chj990707:
To fix the arrows being on the chunk without any blocks, entities will be checked for 1 blocks around them for finding compiled chunks. while this is fine for most case, because projectiles aren't bigger than 1 block most of the time, it might cause problem with big projectiles stuck on wall.

chj990707:
This was rebased to 2.3 and made into another PR.

PR #1326, opened 2025-10-21
Relative movement with ships

chj990707:
Items thrown/dropped on a ship will be dragged accordingly.

Projectiles shot on a ship will have the velocity added instead of being dragged, for ease of other operations.

Dispenser item behaviour was removed because it could be handled by other generic features.

If an entity jumps on a ship and start being dragged, the deltamovement will be changed to relative velocity on the ship.
- Previously this caused the entity to slide across the ship.

At the end of the dragging, either by airborne too long or touching ground, the dragged entity will be pushed with deltaMovement.
- Previously this was just done by attenuating the addedMovement.

VSGameUtils has method called applyShipVelocity for adding ship's velocity to an entity. Main purpose is to set the relative velocity on ship frame to zero vector.

PacketEntityShipMotion, PacketMobShipRotation will only be sent to players who should render the entity.
LivingEntity body rotation will be dragged too.
- Previously only the head was dragged.


ThePlasticPotato (approved):
lgtm!

PR #1327, opened 2025-10-21
Sodium section distance

chj990707:
Sodium render sections(chunks) on the ships was always added at the last of the rebuild list, causing ship rendering delayed until the world chunks submitted are all rendered.
This PR will sort all the chunks by ships and world together by distance. Although sodium field name implies that the rebuild list is sorted at creation, it might have side effect because of the sorting.

Note: Render distance of the ship doesn't match the render distance of the world.


chj990707:
This is moved to 2.3 branch.

PR #1330, opened 2025-10-26
[1.20.1/2.3] Dynmap Support

TechTastic:
This PR brings Dynmap support into VS!

Image of the Dynmap Support in action

As shown above, it adds both an icon marker and an encompassing polyline marker. These are, of course, configurable via the new Dynmap configuration section.

Config

It also adds new compile only dependencies pointing to the Modrinth versions of Dynmap and maintains loader-specific functionality by having a base `DynmapHandler` in common for most of the logic and `FabricDynmapHandler`/`ForgeDynmapHandler` for platform-specific calls. This is necessary due to the differing classpaths and classes available to Forge and Fabric within Dynmap.

Ship markers are updated at the end of gametick from the `allShips` list.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -27,6 +27,9 @@ dependencies {
     // TIS-3d
     modCompileOnly("maven.modrinth:tis3d:${tis3d_version}")
 
+    // Dynmap
+    modCompileOnly("maven.modrinth:dynmap:${dynmap_version}")
+
     //Common create compat,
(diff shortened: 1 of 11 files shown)


TechTastic:
Here is a live demo of it working

http://23.230.3.65:25651/?worldname=world&mapname=surface&zoom=7&x=0&y=64&z=6

blockninja124 (approved):
LGTM

thetheaplant (approved):
Does what it says on the tin

PR #1332, opened 2025-10-29
2.3 rendering chunks sorting

chj990707:
Chunk rebuild order for vanilla and sodium was always last for the chunks on the ships, because their position is in shipyard.
This PR changes the distance to be calculated against ship position, and sorts the chunks accordingly.


chj990707:
Holy shit I rebased wrongfully. I'll get back soon.

PR #1333, opened 2025-10-29
Chunk build sorting on ships

chj990707:
Previously in Vanilla/Sodium Level Renderer chunks on ships were always the last to be built and rendered. This PR changes the distance check for chunks on the ship, and sorts the rebuild list accordingly.

MixinExtras were updated to 0.4.1 to use WrapMethod, and also Fabric loader was updated to follow it.


Rubydesic (approved):
Very nice work!

PR #1336, opened 2025-10-29
skip biome particles in ship chunks

jamesgreen26:
Calling ``thisAsClientLevel.getBiome(mutableBlockPos)`` on every frame for every air block in a ship adds needless overhead. Also, biome particles for ship chunks look strange anyways.


blockninja124 (approved):
Works on both forge and fabric, and makes sense for performance. It also makes sense since the shipyard biome may not be the same as the worldspace biome, which could lead to seeing multiple biome particles which would be weird.

Other particles still work fine on ships.

PR #1337, opened 2025-10-29
Merge create 6 compat onto 2.3

blockninja124:
There are still a few very minor (mostly visual) bugs, but this branch seems basically ready for release so its about time it gets merged into 2.3


blockninja124:
Not going to squash this one, because its waay too many commits to be bundled up in one :P

PR #1339, opened 2025-10-30
Flywheel compat bug fix

chj990707:
From the Create 6 fix PR, the entity visuals(contraptions) on ship could be invisible upon reloading, and visuals could be duplicated on unloading and reloading the ship from distance.

Also, if you load the world for the first time after the game boots, it could cause contraptions to load earlier than the ships, therfore can't have a embedding bound to the ship. To prevent this Ship Embedding Manager will hold a queue of entities which are in shipyard but the ship isn't loaded yet, and enqueue them to the Visualization Manager after the ship is loaded.

Latter needs more investigation because the load order also seems to affect shipyard entities like minecarts, making them not interactable.


chj990707:
Fixed the entity section itself which could be loaded earlier than the ship on client side. Now minecarts don't suffer from non-interactability.

blockninja124:
Works on forge client + server, and fabric client, but unfortunately crashes on fabric dedicated server from trying to access LocalPlayer somewhere. This should probably be fixed before we merge this, although we do need to merge it before release.

PR #1343, opened 2025-10-31
make create air current segment change by other ships

zyxkad:
Based on #1339

- **check passable block for create air current**
- **make create air current segment change by other ships**


blockninja124:
works really well, and can probably be merged after https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1339 is.

PR #1344, opened 2025-11-01
Mass calculation improvements

zaafonin:
This PR makes it possible to specify block masses via "density" parameter, as opposed to "mass". In this case, mass is calculated from density and volume of block's voxelshape. Mass still takes priority over density.

Default values of mass, elasticity, friction and unused hardness are now exposed and tweakable in server side config. The values are also exposed (VSBlockStateInfo class is still private, dw) so in theory we can have values like friction in item tooltips. For now these values are in F3 screen (used to be an Oddities feature), obviously not networked, that's a VS Jade kind of job.

Fluid parameters (density, drag coefficient) are no longer hardcoded and instead can be specified in vs_mass/filename.json files. Right now it uses fluid tags to handle both static and flowing water, behavior with water and lava should be preserved.

In need of testing ;)

https://github.com/user-attachments/assets/f3b7ded0-745d-45a9-9329-632aa4760afe


LogicWheat:
has conflicts
Is this still under development?

PR #1345, opened 2025-11-01
1.20.1/2.3 cherry picked fixes

zaafonin:
Small fixes from 2.5 that don't integrate with mods.


blockninja124:
Works well on both loaders. Note to other testers: if you checkout this PR branch, it will use create 0.5.1, so if you want to test with create 6 rebase this branch onto 1.20.1/2.3 first. There shouldn't be any merge conflicts.

This is good to merge.

PR #1350, opened 2025-11-02
Calculate default mass values from volume and default density

jamesgreen26:
Calculate block masses from voxel shape and density = 1000kg / m3, giving more accurate block mass values.


zaafonin:
This almost looks tidier than mine, I think we should work together on the density feature

jamesgreen26:
Thanks! We totally should work together. I only saw your PR after I posted mine, the customizable depth could be really useful. Especially if it's possible to group blocks of a certain material together

PR #1354, opened 2025-11-02
[1.20.1] [2.4?] Dreams of a Dead Engine

ThePlasticPotato:
haha interloper reference

Objective: Splice together the new API with 2.3 in preparation to return to the original Krunch.


blockninja124:
Here we go again

PR #1361, opened 2025-11-04
Fixed relogs not teleporting players + removed now unused config option

### LINKED ISSUE #486: when you log off on a ship and it moves, you should log on still in the ship, not in the middle of nowhere

m-doescode:
Fixes #486.... Again...

As reported by @blockninja124: https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/486#issuecomment-3486857669, the current solution no longer works. This PR fixes that.

(More than an) edge-case: When a ship "splits" (ie, a joining part breaks, and the two halves are split into two separate ships), there is a chance the player will be standing on the new half, so they will not be teleported. Ideally we'd use a more sophisticated system on the server to store such positions, and automatically translate them, but for right now I think this is good enough.

If you really want me to, I can attempt to tackle that, but I will need pointers as to where that splitting code is, a cursory search didn't lead me anywhere.

Anyway, hope this is helpful for a feature patch update

P.S. Peer review/peer testing appreciated!

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/teleport_reconnected_player_to_ship/MixinServerPlayer.java
@@ -62,7 +62,7 @@ public abstract class MixinServerPlayer extends Player {
     void rememberLastShip(final CompoundTag compoundTag, final CallbackInfo ci) {
         final EntityDraggingInformation draggingInformation = ((IEntityDraggingInformationProvider) this).getDraggingInformation();
 
-        if (draggingInformation.getTicksSinceStoodOnShip() > VSGameConfig.SERVER.getMaxAirborneTicksForReconnectedPlayerTeleport())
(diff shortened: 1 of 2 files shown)


blockninja124:
did getTicksSinceStoodOnShip() just not exist anymore or something? :facepalm:

m-doescode:
> did getTicksSinceStoodOnShip() just not exist anymore or something? 🤦

I'm not sure, but it seems like it no longer reports 0 while the player is standing on the ship

blockninja124:
I've changed the base since in theory 1.20.1/2.4 is going to be the branch that will be worked on going forward, although its still a bit WIP at the moment.

Rubydesic:
lgtm. I don't think it's worth attempting to solve this for ship splits at the moment. Seems more complicated than its worth.

Rubydesic (approved):

m-doescode:
> lgtm. I don't think it's worth attempting to solve this for ship splits at the moment. Seems more complicated than its worth.

I've thought about it, and it doesn't sound that bad. I'd just need to store the information in a public list rather than nbt, then modify it during splitting.

That said, it's definitely going to be something for another pr.

PR #1366, opened 2025-11-05
Remade the in ship fluid push mixin(s)

### LINKED ISSUE #536: Increased fluid resistance when walking

- [X] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

1.16

### Mod Loader

Forge

### Issue description

This bug happens in v2.3.0-beta.3 for 1.20.1. I unfortunately can't select the proper version with this issue template (see #535).

When walking in water or lava, the resistance when walking against the flow is increased compared to vanilla behaviour.
In Vanilla, when walking without jumping or swimming, the player can walk from the water block, which is furthest from the source, all the way to the source block.
With Valkyiren Skies installed, a player can only work the first 3 blocks before beeing pushed back,

The same happens with lava in the Nether. In the overworld, lava seems fine due to its shorter flow distance.

### Issue reproduction

This issue can be reproduced by placing down a water source block in the overworld or a lava source block in the nether.
Then the player can move towards the source block.

### Logs

_No response_

cylian914:
The previous way was overriding the function (and copying vanilla behaviour to apply it to ship) which isn't great for compatibility with other mod that mixin in this function

this uses more complex mixin to let vanilla/forge gather all the fluid that will push the entity then inject code that recall the function and switching variable to make the gathering of ship (reuseing the function make it more compatible with other mod's mixin) and combining with vanilla then letting the vanilla/forge way of managing fluid pushing

i recommend using the mixin.debug.export to understand where it modify and capture value

it also fix #536 and make in ship fluid push in fabric (it wasn't the case before)

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
@@ -201,7 +201,6 @@ private fun getShipObjectManagingPosImpl(world: Level?, chunkX: Int, chunkZ: Int
  * followed by the AABB in the ship-space of the intersecting ships.
  */
 fun Level.transformFromWorldToNearbyShipsAndWorld(aabb: AABB, cb: Consumer<AABB>) {
-    cb.accept(aabb)
     val tmpAABB = AABBd()
     getShipsIntersecting(aabb).forEach { ship ->
         cb.accept(tmpAABB.set(aabb).transform(ship.worldToShip).toMinecraft())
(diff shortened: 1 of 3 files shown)


blockninja124:
I know I said earlier PRs should be made to 2.4, and that is the case going forward, however we are preparing for a final beta release on 2.3 fixing some of the create compat bugs from beta.11. This fix could probably be included. Is there a commit from 2.5 we could cherry pick into 2.3? I presume I can't just rebase this PR into 2.3, it'll get confused. :thinking:

cylian914:
it can 100% be cherry picked for 2.3

#1368

PR #1367, opened 2025-11-05
Add changelog system to auto-publishing

blockninja124:
Also pre-emptively added changelog for beta.12

Probably needs @Rubydesic approval before merging.

Basically, both the curseforge and modrinth publish tasks now read the `changelog.md` file and use it as a (markdown) changelog. I've used this on starlance, so I know it works.


blockninja124:
Interestingly, circle ci is failing. It appears to be looking for the changelog file in a different location than when run on my local repo. (Its looking in /fabric/changelog.md instead of just /changelog.md).

PR #1368, opened 2025-11-05
Remade the in ship fluid push mixin(s)

cylian914:
The previous way was overriding the function (and copying vanilla behaviour to apply it to ship) which isn't great for compatibility with other mod that mixin in this function

this uses more complex mixin to let vanilla/forge gather all the fluid that will push the entity then inject code that recall the function and switching variable to make the gathering of ship (reuseing the function make it more compatible with other mod's mixin) and combining with vanilla then letting the vanilla/forge way of managing fluid pushing

i recommend using the mixin.debug.export to understand where it modify and capture value

it also fix https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/536 and make in ship fluid push in fabric (it wasn't the case before)


blockninja124:
I haven't tested outside the dev enviroment, but inside it works as promised and lgtm.

PR #1369, opened 2025-11-06
fix sodium renderer may crash on custom client level

zyxkad:


blockninja124:
Is this for litematica?

zyxkad:
ah yes

Rubydesic (approved):

PR #1370, opened 2025-11-06
Fix tnt explosion clip

zyxkad:
- **fix explosion when impacted entity is in shipyard**
- **fix typo**

When create contraptions which on a ship are under a TNT explosion, server will get spammed with clip warn


blockninja124:
LGTM

PR #1372, opened 2025-11-06
make fabric/forge dependency update with gradle's

cylian914:
I noticed it wasn't dynamic for some reason

also rearrange the gradle properties to avoid repetition

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -10,7 +10,7 @@ dependencies {
 
     modApi("me.shedaniel.cloth:cloth-config:${cloth_config_version}")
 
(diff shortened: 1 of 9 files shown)


Rubydesic:
Not really convinced this is improvement. Also, just because we have a specific version of a mod in our dev environment doesn't mean we specifically have a hard dependency on that version or higher

cylian914:
if it work on a lower version of a mod then there is no reason to use the a newer one for dev especially since all the testing will be done in the newer version

but if its an actual issue the dynamic version could be put in a recommends and let the break for version that we are sure it will break and update it manually

PR #1376, opened 2025-11-09
Fix mixin not working outside of dev

cylian914:
i though fabric applied the mapping but apparently no


blockninja124:
I think fabric is weird and defaults to not remapping

cylian914:
i just noticed this doesn't compile because i somehow added an additional Inject anotation... SOMEHOW 

its fix anyway so thx

PR #1398, opened 2025-11-19
make it so that the far clip plane is as far as the furthest ship aabb corner

m3t4f1v3:
doesn't seem to affect much, asked around and apparently it MIGHT reduce precision if unload distance is excessively high but like doesn't really matter

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/MixinGameRenderer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/MixinGameRenderer.java
@@ -1,5 +1,6 @@
 package org.valkyrienskies.mod.mixin.client.renderer;
 
+import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.mojang.blaze3d.systems.RenderSystem;
@@ -19,6 +20,7 @@ import org.joml.Quaterniond;
 import org.joml.Quaternionf;
 import org.joml.Vector3d;
 import org.joml.Vector3dc;
+import org.joml.primitives.AABBdc;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/MixinGameRenderer.java
```diff
@@ -295,4 +297,27 @@ private void setupCameraWithMountedShip(final LevelRenderer instance, final Pose
             this.getProjectionMatrix(Math.max(fov, this.minecraft.options.fov().get())));
     }
     // endregion
+
+    @WrapMethod(method = "getDepthFar")
+    public float includeShipsIn(Operation<Float> original) {
```
Rubydesic: Can you use [`@ModifyReturnValue`](https://github.com/LlamaLad7/MixinExtras/wiki/ModifyReturnValue) instead?

common/src/main/java/org/valkyrienskies/mod/mixin/client/renderer/MixinGameRenderer.java
```diff
@@ -295,4 +297,27 @@ private void setupCameraWithMountedShip(final LevelRenderer instance, final Pose
...
+            double furthestDistanceSq = 0;
+            for (int i = 0; i < 8; i++) {
+                double x = (i & 1) == 0 ? shipAABB.minX() : shipAABB.maxX();
+                double y = (i & 2) == 0 ? shipAABB.minY() : shipAABB.maxY();
+                double z = (i & 4) == 0 ? shipAABB.minZ() : shipAABB.maxZ();
+                double distanceSq = cameraPos.distanceToSqr(new Vec3(x, y, z));
+                if (distanceSq > furthestDistanceSq) {
+                    furthestDistanceSq = distanceSq;
+                }
+            }
```
Rubydesic: Can we replace the loop with a simpler calculation, like this?

```java
double dMinX = shipAABB.minX() - cameraPos.x();
double dMaxX = shipAABB.maxX() - cameraPos.x();
double dMinY = shipAABB.minY() - cameraPos.y();
double dMaxY = shipAABB.maxY() - cameraPos.y();
double dMinZ = shipAABB.minZ() - cameraPos.z();
double dMaxZ = shipAABB.maxZ() - cameraPos.z();
double furthestDistSq = Math.max(dMinX * dMinX, dMaxX * dMaxX) + Math.max(dMinY * dMinY, dMaxY * dMaxY) + Math.max(dMinZ * dMinZ, dMaxZ * dMaxZ);
maxDistance = Math.max(maxDistance, (float) Math.sqrt(furthestDistSq));
```

Rubydesic (changes requested):

Rubydesic (approved):
LGTM!

PR #1402, opened 2025-11-21
Re-orginized weights

blockninja124:
Moved all of the weights into (mostly) organized files in folders that loosely follow the creative inventory tabs. While its far from perfect, I think its way better than it was before. (with some vague organization, and then just 1_18 1_19 1_20 files for some reason). 

Looking for feedback on the changes, and maybe some suggested changes for specific weights, so I'll leave this as a draft for now.


blockninja124:
Eh its good enough to merge as is

jamesgreen26:
Merged a duplicate PR

PR #1412, opened 2025-11-24
Fix shipyard entity rendering when pehkui installed

jamesgreen26:
Change mixin priority so that VS transforms rendering to the shipyard *before* other mods do their rendering shenanigans, which produces correct visuals.

Before:
Screenshot_2025-11-23_at_6 17 37_PM

After:
Screenshot 2025-11-23 at 6 46 16 PM


blockninja124:
Goated.

Here we were thinking it needed complex compat rendering shenanigans :facepalm:

jamesgreen26:
https://github.com/user-attachments/assets/7bb51985-2dbf-4cec-82b2-b0b023015c29

PR #1418, opened 2025-11-26
Fixed game crashing without Sodium installed when ships are present

zaafonin:
SodiumCompat checks if Sodium is installed which should be enough to prevent crashes but actually isn't, merely calling anything from SodiumCompat crashes the game because it references Sodium classes 🤷‍♂️ duct taping over that, it's used only these three times anyway


ThePlasticPotato:
high priority, merging so i hope this works

PR #1420, opened 2025-11-28
Schematicannon rendering fix, better mixin targets for BE

zaafonin:


blockninja124:
Looks good, but I think toSameSpaceAs needs some javadoc (or kotlin doc? idk) explaining what it does at first glance. Also an explanation of why there are so many overloads (because its likely to be called from java mixins a lot), so someone else doesn't think its excessive when VectorConversions exists. I also think we should ask Ruby if we can move it to VSGameUtils so that we don't have this random CompatUtils class with only one method (even though that is what we had before).

I also suspect once this is merged, a lot of our previous mixins can switch to using toSameSpaceAs, but thats for later.

blockninja124 (approved):
Looks good to me

StewStrong:
Thx for fixing this :)

common/src/main/kotlin/org/valkyrienskies/mod/common/CompatUtil.kt
```diff
@@ -1,41 +1,97 @@
...
+    // Same method is repeated several times with different argument types, as nearly all of our compat code is in Java
+    // which lacks type conversion sugar of Kotlin. This makes for cleaner code in mixins themselves because we avoid
+    // many usages of VectorConversionUtilsKt just to deal with JOML types used in VS.
+
+    /**
+     * Transform an arbitrary position to shipspace of a set ship.
+     * This method handles coordinates in worldspace, shipspace of other ships and shipspace of the same ship.
+     */
+    fun toSameSpaceAs(level: Level?, position: Vector3dc, targetShip: Ship?): Vector3d {
+        val result = position as Vector3d
```
Rubydesic, line 26: what the fuck

StewStrong: Fixed in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/2bc7b5ea96ecd53d58dfb822a913e9da365b626d

PR #1422, opened 2025-11-28
Actually use the runRegisterBlockStateEvent method

blockninja124:
Now the registerBlockStateEvent event is _actually_ emitted


thetheaplant (approved):
yeah sure does that

PR #1450, opened 2025-12-08
Basic controller integration

thetheaplant:
Very basic, thank you getitem for the help and tweaked controllers reference. Needs extensive testing, not sure if i broke anything

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSKeyBindings.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSKeyBindings.kt
@@ -38,3 +38,6 @@ object VSKeyBindings {
         return keyMapping.name.startsWith("key.valkyrienskies")
     }
 }
+    val shipUp = KeyMapping("key.vs.ship_up", InputConstants.Type.KEYSYM, GLFW.GLFW_GAMEPAD_BUTTON_A, "key.categories.vs")
+    val shipDown = KeyMapping("key.vs.ship_down", InputConstants.Type.KEYSYM, GLFW.GLFW_GAMEPAD_BUTTON_B, "key.categories.vs")
+}
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSKeyBindings.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSKeyBindings.kt
@@ -4,6 +4,7 @@ import net.minecraft.client.KeyMapping
 import org.lwjgl.glfw.GLFW
 import java.util.function.Consumer
 import java.util.function.Supplier
+import com.mojang.blaze3d.platform.InputConstants;
 
 object VSKeyBindings {
     // TODO when making the addon utils for registering... this too
@@ -38,6 +39,7 @@ object VSKeyBindings {
         return keyMapping.name.startsWith("key.valkyrienskies")
     }
 }
+object VSControllerBindings {
     val shipUp = KeyMapping("key.vs.ship_up", InputConstants.Type.KEYSYM, GLFW.GLFW_GAMEPAD_BUTTON_A, "key.categories.vs")
     val shipDown = KeyMapping("key.vs.ship_down", InputConstants.Type.KEYSYM, GLFW.GLFW_GAMEPAD_BUTTON_B, "key.categories.vs")
 }
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
@@ -145,6 +145,7 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
         val up = isPressed(opts.keyJump, GLFW.GLFW_GAMEPAD_BUTTON_A)
         val down = isPressed(VSKeyBindings.shipDown.get(), GLFW.GLFW_GAMEPAD_BUTTON_B)
         val cruise = isPressed(VSKeyBindings.shipCruise.get(), GLFW.GLFW_GAMEPAD_BUTTON_Y)
+        // can't believe no one uses this
         val sprint = this.controllingPassenger?.isSprinting == true
 
         val leftStickX = if (selectedGamepad >= 0) state.axes(GLFW.GLFW_GAMEPAD_AXIS_LEFT_X) else 0f
@@ -162,8 +163,8 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
 
         val magnitude = impulse.length()
         if (magnitude > 1f) impulse.mul(1f / magnitude)
-        
-        vsCore.simplePacketNetworking.run {
+
+        with(vsCore.simplePacketNetworking) {
             PacketPlayerDriving(impulse, sprint, cruise).sendToServer()
         }
     }
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
@@ -82,25 +82,85 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
         super.remove(removalReason)
     }
 
+    // basic controller compat, mostly from Tweaked Controllers and Controllable, from GetItemFromBlock and Mrcrawfish, respectively
+    private var selectedGamepad = -1
+    private val state = GLFWGamepadState.create()
+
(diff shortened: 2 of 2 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
...
+                    uniqueGamepadID = -2
+                }
+                GLFW.glfwGetGamepadState(i, state)
+                for (b in 0..14) {
+                    if (state.buttons(b).toInt() != GLFW.GLFW_PRESS) continue
+                    selectedGamepad = i
+                    break
+                }
+            }
+            // ig we need to do this
```
getItemFromBlock, line 110: The whole point of the first part of this function is to try and detect the case where only a single gamepad is connected, in which case you can use that instead of looking for an active one (which is what the few lines above do)
The `uniqueGamepadID` variable is initialized at -1, stores a value >= 0 if there is only a single gamepad connected and -2 if mutiple gamepads were found.
Here we just check that the gamepad was not already found/assigned and that we found a single one.

common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSKeyBindings.kt
```diff
@@ -38,3 +39,7 @@ object VSKeyBindings {
         return keyMapping.name.startsWith("key.valkyrienskies")
     }
 }
+object VSControllerBindings {
+    val shipUp = KeyMapping("key.vs.ship_up", InputConstants.Type.KEYSYM, GLFW.GLFW_GAMEPAD_BUTTON_A, "key.categories.vs")
```
getItemFromBlock, line 43: Interesting, I didn't know minecraft bindings could take gamepad inputs like that
But I am worried that you cannot bind them to a gamepad via the in-game menu, as far as I know the game does not listen for that and only expect mouse/keyboard keys.

getItemFromBlock: Wait this is actually unused, you might want to remove it or put a comment. I suspect this was an attempt at making them bindable via the in game menu?

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
getItemFromBlock: The problem with this code is that it only allows a single gamepad, AND it only allows something that GLFW _thinks_ is a gamepad. For most users this will be fine, but as you can see by the tweaked controllers bug reports it does not fit all cases.
Making it work with more exotic controllers and with multiple controllers is a much much harder task, and would be its own PR/integration on its own. Leaving the code as is is perfectly fine for this PR imo.

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
         super.remove(removalReason)
     }
 
+    // basic controller compat, mostly from Tweaked Controllers and Controllable, from GetItemFromBlock and Mrcrawfish, respectively
+    private var selectedGamepad = -1
+    private val state = GLFWGamepadState.create()
```
getItemFromBlock, line 90: Sidenote about this: there is an underlying c struct behind this variable that is dynamically allocated, hence why it needs to be created with this specific function.
Having it as a "global" variable here is good as it means the struct will be allocated only once at startup

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
...
+        // praise be getitemfromblock
+        fun isPressed(binding: KeyMapping?, buttonIndex: Int?): Boolean {
+            val kb = binding?.isDown ?: false
+            val gp = buttonIndex?.let {
+                selectedGamepad >= 0 && state.buttons(it).toInt() == GLFW.GLFW_PRESS
+            } ?: false
+            return kb || gp
+        }
+
+        // control deez nuts
```
getItemFromBlock, line 140: nuts indeed

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
...
+        // can't believe no one uses this
         val sprint = this.controllingPassenger?.isSprinting == true
-        val down = VSKeyBindings.shipDown.get().isDown
-        val cruise = VSKeyBindings.shipCruise.get().isDown
+
+        val leftStickX = if (selectedGamepad >= 0) state.axes(GLFW.GLFW_GAMEPAD_AXIS_LEFT_X) else 0f
+        val leftStickY = if (selectedGamepad >= 0) state.axes(GLFW.GLFW_GAMEPAD_AXIS_LEFT_Y) else 0f
+        val rightStickY = if (selectedGamepad >= 0) state.axes(GLFW.GLFW_GAMEPAD_AXIS_RIGHT_Y) else 0f
+
+        // stops ship helms from constantly turning, can probably be axed if i make full analog support later
```
getItemFromBlock, line 155: Deadzones are such a PITA to deal with, some gamepads will not have this problem, some will have so much that even values such as 0.2 are too small
Ideally the deadzone should be circular (aka taking the length of the XY vector and comparing its length) but making it square is good enough
Also ideally it should be configurable, but this can be easily done in a later PR

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
...
+        fun applyDeadzone(value: Float, deadzone: Float = 0.1f) =
+            if (kotlin.math.abs(value) < deadzone) 0f else value
 
         val impulse = Vector3f()
-        impulse.z = if (forward == backward) 0.0f else if (forward) 1.0f else -1.0f
-        impulse.x = if (left == right) 0.0f else if (left) 1.0f else -1.0f
-        impulse.y = if (up == down) 0.0f else if (up) 1.0f else -1.0f
+        impulse.x = applyDeadzone(-leftStickX) + if (left == right) 0f else if (left) 1f else -1f
```
getItemFromBlock, line 160: the original code was already a bit confusing, but now with the added condition and the lack of parenthesis its even worse
I suggest rewriting it to something like that:
```
impulse.x = applyDeadzone(-leftStickX)
if (left) impulse.x += 1.0f
if (right) impulse.x -= 1.0f

// do the same for y and z axis...
```
If both left and right are pressed, the last two lines cancels out.
Looks more readable imo, but do note that I am not experienced enough with kotlin. Maybe single liners like that are more common?

common/src/main/kotlin/org/valkyrienskies/mod/common/entity/ShipMountingEntity.kt
```diff
@@ -82,23 +85,84 @@ open class ShipMountingEntity(type: EntityType<ShipMountingEntity>, level: Level
...
+        impulse.z = applyDeadzone(-leftStickY) + if (forward == backward) 0f else if (forward) 1f else -1f
+        impulse.y = applyDeadzone(-rightStickY) + if (up == down) 0f else if (up) 1f else -1f
+
+        val magnitude = impulse.length()
+        if (magnitude > 1f) impulse.mul(1f / magnitude)
```
getItemFromBlock, line 165: Interestingly, the previous code didn't normalize the vector like that allowing for values like `Vector3f(1,0,1)` to be accepted.
I don't know if this will have any impact on the game, but it might make it harder to turn while moving forward.
I'd recommend doing a element-wise clamp instead just in case so that the original behavior will be unchanged:
```
        fun clamp(value: Float): Float {
            if (value > 1f) value = 1f
            else if (value < -1f) value = -1f
            return value
        }
        impulse.x = clamp(impulse.x)
        impulse.y = clamp(impulse.y)
        impulse.z = clamp(impulse.z)
```

getItemFromBlock (approved):
Good first implementation, will actually test tomorrow because eepy but by looking at the code everything seems to work
I tried to include useful comments among things that I feel would benefit from being changed a little bit.
I did the review in github directly, the code snippets might not compile.

PR #1456, opened 2025-12-10
Fixed `ValkyrienSkies#distanceSquared` to support World <-> Ship

TechTastic:
Previously this method only transformed positions when both positions were on separate Ships, now it will do it if either is on a Ship or both are on separate Ships


blockninja124 (approved):
lgtm, weird it didn't do this before.

Rubydesic (approved):
not sure what kinda crack I was smoking when I wrote this

PR #1458, opened 2025-12-10
Add basic assembly blacklist addons can use

blockninja124:
Adds an `assemble_blacklist` block tag, registered as a `TagKey` in `ValkyrienSkiesMod`, along with a `BlockState` extension function in `VSGameUtilsKt`.
```kotlin
/**
 * Is the [BlockState] in the `assemble_blacklist`
 * ([ValkyrienSkiesMod.ASSEMBLE_BLACKLIST]) block tag
 */
fun BlockState?.inAssemblyBlacklist(): Boolean {
    return this?.`is`(ASSEMBLE_BLACKLIST) ?: false
}
```

After some deliberation, the choice was made to **not** include natural blocks in the blacklist. This is because it would encourage addons to use the blacklist (which would inevitable fail with modded natural blocks), instead of adding proper anti-assemble-entire-world checks.

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
@@ -1,7 +1,6 @@
 package org.valkyrienskies.mod.common
 
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation
-import com.mojang.logging.LogUtils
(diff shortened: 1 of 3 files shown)


TechTastic (approved):
LGTM! Ive already pointed a few developers of other mods towards this

jamesgreen26:
Looks great!

jamesgreen26 (approved):

PR #1460, opened 2025-12-12
Update mass when default mass is changed in config

jamesgreen26:
Also implement an event for when config is updated


jamesgreen26:
https://github.com/user-attachments/assets/a2052b7a-566b-4918-b9f1-4e2f4bb06372

PR #1466, opened 2025-12-13
fix sneaking on ships

jamesgreen26:
Now sneaking on a ship works as it should, where you can walk at a slower speed without falling off the edge of the ship.

Limitations: sometimes it doesn't let you walk while shifting if you're right in the corner of a block, or standing on something narrow like a fence post


jamesgreen26:
Just realized that Tomato had fixed this earlier, since then something had broken it

jamesgreen26:
https://github.com/user-attachments/assets/93ab4c00-7944-455a-95e7-2895883236b8

LogicWheat:
this sneak is better, i love it

StewStrong:
Ok, this is pretty based!

ThePlasticPotato:
Fixed the original stuff!

PR #1472, opened 2025-12-17
[1.20.1] Gradle Cleanup - Less Transitive Dependencies and Repository Content Exclusion

TechTastic:
So, to lower the burden of depending on VS in other projects and to speed up the build time of VS itself, I have refactored, removed, and improved Gradle dependencies and repostories

# Transitive Cleanup
So, the hardest part of depending on VS is the over abundance of transitive dependencies it brings with it. Not every addon or project needs Create pulled from VS nor VMP.

So, to clean this up, I swapped `modImplementation` dependencies for a combination of `modCompileOnly` and `modRuntimeOnly`. This makes such dependencies non-transitive and thus will not show up in projects depending on VS.

I also removed dependencies we no longer need, like Cloth Config, and duplicates, like Weather 2.

# Repository Content Exclusion
Another issue is the build time of VS, which is not helped when all the dependencies are trying to pull from *every* repository. To prevent this, I have acced `content { }` blocks to most, if not all, repositories so searching for the correct repository is ***much*** faster.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,18 +97,26 @@ subprojects {
             }
             maven {
                 url = "https://cursemaven.com"
-                content { includeGroup("curse.maven") }
+                content {
+                    includeGroup("curse.maven")
+                }
             }
-            maven { // Mod Menu
+            maven {
+                name = "Mod Menu"
                 url = "https://maven.terraformersmc.com/releases/"
-                content { includeGroup("com.terraformersmc") }
+                content {
+                    includeGroup("com.terraformersmc")
+                }
             }
             maven {
                 name = 'Kotlin for Forge'
                 url = 'https://thedarkcolour.github.io/KotlinForForge/'
-                content { includeGroup("thedarkcolour") }
+                content {
+                    includeGroup("thedarkcolour")
+                }
             }
             maven {
+                name = "Registrate"
                 name = 'tterrag maven'
                 url = 'https://maven.tterrag.com/'
                 content {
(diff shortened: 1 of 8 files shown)


thetheaplant (dismissed):
probably fine

build.gradle
```diff
@@ -90,44 +90,103 @@ subprojects {
...
             mavenLocal()
             maven {
                 name = "ParchmentMC"
                 url = "https://maven.parchmentmc.org"
             }
-            maven { url = "https://cursemaven.com" }
-            maven { url = "https://maven.terraformersmc.com/releases/" } // Mod Menu
+            maven {
+                url = "https://cursemaven.com"
+                content { includeGroup("curse.maven") }
```
zyxkad: I do not prefer compressed style here, since it make desync `content` block style.

build.gradle
```diff
@@ -90,44 +90,103 @@ subprojects {
...
             }
             maven {
                 name = 'tterrag maven'
                 url = 'https://maven.tterrag.com/'
+                content {
+                    includeGroup("com.tterrag.registrate")
+                    includeGroup("com.tterrag.registrate_fabric")
+                }
+            }
+            maven { // Twilight Forest
```
zyxkad: Instead of adding the name comment, gradle has the `name = "XXX"` field, which will also be written in the published files.

common/build.gradle
```diff
@@ -1,26 +1,26 @@
 
 dependencies {
-    implementation(annotationProcessor("io.github.llamalad7:mixinextras-common:0.4.1"))
+    implementation(annotationProcessor("io.github.llamalad7:mixinextras-common:$mixinextras_version"))
```
zyxkad, line 3: IMO `${var}` is better than `$var`, because when there it is easier to identify where the variable name starts and end with the bracket pairs, especially when the variable is in a middle of a long string.
In addition, use brackets can prevent confusing between things such as `${var}_` and `$var_`.

TechTastic: I would agree looking at it from Github, but in IntelliJ, the inline variables are highlighted, either in a light purple or white compared to the cactus green of the string.

zyxkad: I don't get you, if the highlighting is annoyed I think you can just modify the settings?

build.gradle
```diff
@@ -97,67 +97,94 @@ subprojects {
...
+                name = "Mod Menu"
                 url = "https://maven.terraformersmc.com/releases/"
-                content { includeGroup("com.terraformersmc") }
+                content {
+                    includeGroup("com.terraformersmc")
+                }
             }
             maven {
                 name = 'Kotlin for Forge'
                 url = 'https://thedarkcolour.github.io/KotlinForForge/'
```
zyxkad, line 113: Meanwhile, I think we can unify single quotes or double quotes.

build.gradle
```diff
@@ -142,19 +243,28 @@ subprojects {
...
-            maven { // Hexal
+            maven {
+                name = "Hexal"
                 url = uri("https://maven.hexxy.media")
+                content {
+                    includeGroup("ram.talia.hexal")
+                }
             }
         }
         maven {
```
zyxkad, line 270: I'd also suggest put VS maven before `if (!project.block_external_repositories)`, so if a dependency is loaded in the VS maven it can be used first.

common/build.gradle
TechTastic: You misunderstand. Im saying I think its fine explicitly because of highlighting, it just looks worse from Github.

build.gradle
TechTastic: Sure

build.gradle
TechTastic: Fair enough

TechTastic:
This PR fumbled

Issue #1302, opened 2025-09-14
Concurrent Modification Exception when ship with multi-block accumulator moves to another dimesion


Drnagol369:
### Mod Name

Create: Crafts and Additions, Starlance

### This issue occurs when only Valkyrien Skies, addons, and the mod I have specified are installed and no other mods

- [x] I have tested this issue and it occurs with only Valkyrien Skies, addons, and the mod I have specified

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

When a ship with an accumulator made of more than one block moves to another dimension (e.g. the solar system dimension in Cosmic Horizons) the game crashes. Furthermore, when relogging, and teleporting my ship back to the ground, it had corrupted, with two chunks seemingly erased.

<img width="1050" height="1000" alt="Image" src="https://github.com/user-attachments/assets/a130d34d-4c47-408b-a252-23c14614d410" />
//aftermath of the crash

Ideally this does not occur, with the ship simply travelling to the new dimension.

I have posted this issue on both the Create: Crafts and Additions and Starlance issue trackers, too, as I am unsure as to which mod is most responsible for fixing this bug.

### Issue reproduction

I initially struggled to identify the exact reason for this. After using NCME, however, it was relatively easy to identify the root cause being the accumulator, given how many times its name popped up in the crash log. I then performed some experiments, however could not reproduce it initially, as I did not realise it was for multi-block accumulators only.

<img width="798" height="578" alt="Image" src="https://github.com/user-attachments/assets/1832d4c5-b9e3-4557-82de-dd5b57c7fa55" />
//the four test vehicles, all of which caused a crash

After performing 4 test launches of the electronics of my whole rocket, gradually removing parts on each test, all 4 failed. Only once I created another, which only had the single block accumulator, did the game not crash.

STEPS FOR REPRODUCTION:

 - This is relatively simple to do within Starlance, using a normal thruster on the base, and then either placing the double accumulator on top directly, or placing a chunk loader (which means I don't have to ride in the vehicle itself in order to test). 
 - After assembling, and powering the thruster using a lever, wait for the rocket to reach the altitude at which the dimension will transition over
 - Crash!

### Logs

[crash-2025-09-14_12.52.22-server.txt](https://github.com/user-attachments/files/22320771/crash-2025-09-14_12.52.22-server.txt)

[2025-09-14-9.log](https://github.com/user-attachments/files/22320794/2025-09-14-9.log)

[...]

PR #1473, opened 2025-12-17
fix CME when teleporting ships cross dimension

jamesgreen26:
### Bug:
Teleporting a ship cross dimension will throw a Concurrent Modification Exception if certain blocks are present on the ship. This makes certain mods like Mekanism unusable with mods like Starlance or Genesis.

### Examples:
https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/1302
https://github.com/jcm236/Starlance/issues/9

### Cause:
LevelChunk.clearAllBlockEntities() is not thread safe and we are calling it from a thread-unsafe context while teleporting ships between dimensions.

### Fix:
Rewrite LevelChunk.clearAllBlockEntities() to be thread safe via mixins.


blockninja124:
holy crap this might fix all the cme issues

common/src/main/resources/valkyrienskies-common.accesswidener
```diff
@@ -1,4 +1,4 @@
-accessWidener	v1	named
+accessWidener	v2	named
```
blockninja124, line 1: is updating the access widener required? I worry it might break other stuff

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
```diff
@@ -0,0 +1,65 @@
...
+    @Final
+    private static TickingBlockEntity NULL_TICKER;
+
+    @Inject(method = "clearAllBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void clearAllBlockEntities(CallbackInfo ci) {
+        if (!this.blockEntities.isEmpty()) {
+            List<BlockEntity> blockEntitiesSnapshot = List.copyOf(this.blockEntities.values());
+
+            for (BlockEntity be : blockEntitiesSnapshot) {
+                be.setRemoved();
```
blockninja124, line 46: have you checked to see if this causes blocks like copycats and storage drawers to drop their items? I feel like I remember them doing that a while ago during assembly.

forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
```diff
@@ -0,0 +1,69 @@
...
+    @Final
+    private static TickingBlockEntity NULL_TICKER;
+
+    @Inject(method = "clearAllBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void clearAllBlockEntities(CallbackInfo ci) {
+        if (!this.blockEntities.isEmpty()) {
+            List<BlockEntity> blockEntitiesSnapshot = List.copyOf(this.blockEntities.values());
+
+            for (BlockEntity be : blockEntitiesSnapshot) {
+                be.onChunkUnloaded();
```
blockninja124, line 46: How come this is only on forge? Aside from it, I believe this mixin could be in common :thinking:

blockninja124 (approved):

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
zyxkad: > have you checked to see if this causes blocks like copycats and storage drawers to drop their items? I feel like I remember them doing that a while ago during assembly.

No, the resource drops happens on `Block.onRemove`.

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
```diff
@@ -0,0 +1,65 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(value = LevelChunk.class, priority = 1500)
+public abstract class MixinLevelChunk extends ChunkAccess {
+
+    public MixinLevelChunk(ChunkPos arg, UpgradeData arg2, LevelHeightAccessor arg3, Registry<Biome> arg4, long l, @Nullable LevelChunkSection[] args, @Nullable BlendingData arg5) {
+        super(arg, arg2, arg3, arg4, l, args, arg5);
```
zyxkad, line 29: we do not initialize mixin class ever, so we can just pass a bunch of nulls instead of writing the full signature out.

```suggestion
    public MixinLevelChunk() {
        super(null, null, null, null, 0, null, null);
```

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
```diff
@@ -0,0 +1,65 @@
...
+    @Inject(method = "clearAllBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void clearAllBlockEntities(CallbackInfo ci) {
+        if (!this.blockEntities.isEmpty()) {
+            List<BlockEntity> blockEntitiesSnapshot = List.copyOf(this.blockEntities.values());
+
+            for (BlockEntity be : blockEntitiesSnapshot) {
+                be.setRemoved();
+            }
+
+            this.blockEntities.clear();
```
zyxkad, line 49: I doubt if any BlockEntities is added during `setRemoved`, which is a very rare case yes, they won't be marked as removed but direct get dropped off.
Maybe can let clear happens right after made the copy.

common/src/main/resources/valkyrienskies-common.accesswidener
jamesgreen26: Yes, updating the access widener is required for targeting private inner classes.

Updating this should not have breaking changes. The v2 format was introduced in 2021 and the loom version we are using fully supports it.

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
jamesgreen26: I have not tested with copycats or storage drawers. I assumed it would not make a difference, since I am calling the same methods as vanilla, just without risk of CME. This should probably be tested anyways though.

forge/src/main/java/org/valkyrienskies/mod/forge/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
jamesgreen26: This method does not exist in vanilla/fabric, forge patches it in

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
jamesgreen26: With respect, I don't see the benefit of this change

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
jamesgreen26: I'm not sure which would be better. My reasoning for the current implementation is to match the preexisting behaviour as closely as possible

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
zyxkad: > With respect, I don't see the benefit of this change

You don't need import bunch of classes that is unused.

jamesgreen26: I see. More imports is unproblematic, modern IDEs will collapse the imports section and manage imports automatically.

I will decline this change to stay consistent with existing codebase conventions.

Rubydesic:
Where are we calling LevelChunk.clearAllBlockEntities in an unsafe context? Shouldn't we call it on the right thread instead of patching this method?

zyxkad:
> Where are we calling LevelChunk.clearAllBlockEntities in an unsafe context? Shouldn't we call it on the right thread instead of patching this method?

He probably means some block entity will modify `blockEntities` in `setRemoved`, and / or `clearAllBlockEntities` is get invoked while iterating over the block entities ticker. Not really a thread issue.

jamesgreen26:
@Rubydesic I don't think I described the problem very clearly.  Below I've explained by best understanding of why the crash happens, take it with a grain of salt.

In vanilla, LevelChunk.clearAllBlockEntities() is only called when chunks are unloading, and afterwards the block entities do not persist or keep ticking.

In our usage, the block entities are fist reassigned to another dimension, then we call LevelChunk.clearAllBlockEntities() while the block entities continue to tick. This is not an issue for most block entities but there are a few mods for which this causes problems, such as Mekanism and Create: Crafts&Additions. I don't know which of the hash maps is getting mutated and throwing the CME, but it could be any of them.

A potential solution would be to clear the block entities in the old dimension from the thread in the new dimension? I'm not sure if that would work.
I think the proposed changes in this PR are a better solution since I'm much more confident about how it fixes the crash, and keeps our teleportation code from becoming more complex with asynchronous tasks. However, if anyone has ideas for a better way to fix it I'd be happy to hear about them.

jamesgreen26:
I don't know if the issue is related to threads or something else

zyxkad:
> A potential solution would be to clear the block entities in the old dimension from the thread in the new dimension

Note in Minecraft, all dimensions always ticking on the same server thread unless you installed one of the "optimization" mod.

> and keeps our teleportation code from becoming more complex with asynchronous tasks.

Also please note the `List.copyOf`s are not thread safe here, so do not assume it can work well on multi thread envrionment.

zyxkad:
> I don't know if the issue is related to threads or something else

If you can upload an example crash report, the issue can be analyzed better.

jamesgreen26:
<details>
  <summary>Crashlog</summary>
  
```
// Surprise! Haha. Well, this is awkward.

Time: 2025-12-16 22:31:54
Description: Exception ticking world

java.util.ConcurrentModificationException: null
	at java.util.HashMap$Values.forEach(Unknown Source) ~[?:?] {}
	at net.minecraft.world.level.chunk.LevelChunk.m_187957_(LevelChunk.java:576) ~[client-1.20.1-20230612.114412-srg.jar%23219!/:?] {re:mixin,re:classloading,pl:mixin:APP:fabric-data-attachment-api-v1.mixins.json:WorldChunkMixin,pl:mixin:APP:fabric-lifecycle-events-v1.client.mixins.json:WorldChunkMixin,pl:mixin:APP:copycats-common.mixins.json:foundation.copycat.migration.LevelChunkMixin,pl:mixin:APP:flywheel.impl.mixins.json:visualmanage.LevelChunkMixin,pl:mixin:APP:valkyrienskies-common.mixins.json:feature.physics_block_entities.MixinLevelChunk,pl:mixin:APP:valkyrienskies-common.mixins.json:world.chunk.MixinLevelChunk,pl:mixin:APP:vs_clockwork-common.mixins.json:MixinBlockStateInfoProvider,pl:mixin:A}
	at net.minecraft.world.level.chunk.LevelChunk.clearChunk(LevelChunk.java:3395) ~[client-1.20.1-20230612.114412-srg.jar%23219!/:?] {re:mixin,re:classloading,pl:mixin:APP:fabric-data-attachment-api-v1.mixins.json:WorldChunkMixin,pl:mixin:APP:fabric-lifecycle-events-v1.client.mixins.json:WorldChunkMixin,pl:mixin:APP:copycats-common.mixins.json:foundation.copycat.migration.LevelChunkMixin,pl:mixin:APP:flywheel.impl.mixins.json:visualmanage.LevelChunkMixin,pl:mixin:APP:valkyrienskies-common.mixins.json:feature.physics_block_entities.MixinLevelChunk,pl:mixin:APP:valkyrienskies-common.mixins.json:world.chunk.MixinLevelChunk,pl:mixin:APP:vs_clockwork-common.mixins.json:MixinBlockStateInfoProvider,pl:mixin:A}
	at net.minecraft.server.MinecraftServer.mdfa6de4$lambda$moveTerrainAcrossDimensions$4$1(MinecraftServer.java:4794) ~[client-1.20.1-20230612.114412-srg.jar%23219!/:?] {re:mixin,pl:accesstransformer:B,re:classloading,pl:accesstransformer:B,pl:mixin:A}
	at org.valkyrienskies.core.impl.chunk_tracking.ShipActiveChunksSet.forEach(Unknown Source) ~[valkyrienskies-120-forge-2.4.0+9578397ff3.jar%23213!/:?] {re:classloading}
	at net.minecraft.server.MinecraftServer.moveTerrainAcrossDimensions(MinecraftServer.java:4792) ~[client-1.20.1-20230612.114412-srg.jar%23219!/:?] {re:mixin,pl:accesstransformer:B,re:classloading,pl:accesstransformer:B,pl:mixin:A}
	at org.valkyrienskies.core.impl.shadow.Er.teleportShip(Unknown Source) ~[valkyrienskies-120-forge-2.4.0+9578397ff3.jar%23213!/:?] {re:classloading}
	at shipwrights.genesis.teleportation.TeleportationHandler.handleShipTeleport(TeleportationHandler.java:361) ~[genesis-1.20.1-0.4.0.jar%23200!/:?] {re:classloading}
	at it.unimi.dsi.fastutil.longs.Long2ObjectMap.lambda$forEach$0(Long2ObjectMap.java:249) ~[fastutil-8.5.9.jar%23105!/:?] {re:mixin}
	at it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap$MapEntrySet.fastForEach(Long2ObjectOpenHashMap.java:1027) ~[fastutil-8.5.9.jar%23105!/:?] {}
	at it.unimi.dsi.fastutil.longs.Long2ObjectMap.forEach(Long2ObjectMap.java:251) ~[fastutil-8.5.9.jar%23105!/:?] {re:mixin}
	at shipwrights.genesis.teleportation.TeleportationHandler.finalizeTeleport(TeleportationHandler.java:261) ~[genesis-1.20.1-0.4.0.jar%23200!/:?] {re:classloading}
	at shipwrights.genesis.event.AtmosphericCollision.lambda$atmosphericCollisionTick$2(AtmosphericCollision.java:99) ~[genesis-1.20.1-0.4.0.jar%23200!/:?] {re:classloading}
	at net.minecraft.server.level.ServerLevel.mdfa6de4$lambda$vlib$tick$0$9(ServerLevel.java:5652)
```
</details>

zyxkad:
Okay well, https://github.com/mrh0/createaddition/issues/893 is probably a better example.

What happened here is create addition invoked `getBlockEntity` inside `setRemove` (indirectly). And then Minecraft's `getBlockEntity` will invoke `this.blockEntities.remove` if the block entity is marked as removed. So the logic can be seen as the following:

```java
chunk.blockEntities.forEach((be) -> {
  be.setRemoved(); {
    if (!isLoaded(pos)) return; // short circuit when it is normal unloading
    level.getBlockEntity(pos); {
      chunk.getBlockEntity(pos); {
        be2 = chunk.blockEntities.get(pos);
        if (be2.isRemoved()) {
          chunk.blockEntities.remove(pos); // throws concurrent modification exception
        }
      }
    }
  }
});
```
and as we can expect, `chunk.blockEntities.forEach((be) -> chunk.blockEntities.remove)` will absolutely throws CME.

It works normally when unloading because `isLoaded` will short circuit the method.

jamesgreen26:
Thanks for looking into this @zyxkad , that was super helpful.

I've also found similar logic in Mekanism:

[TileEntityTransmitter](https://github.com/mekanism/Mekanism/blob/41de98670ef19b21c21de67d7198ba1731fe0b5c/src/main/java/mekanism/common/tile/transmitter/TileEntityTransmitter.java#L161)

```
    @Override
    public void setRemoved() {
        super.setRemoved();
        onWorldSeparate(false);
        getTransmitter().remove();
    }
```

It is clear that the crash comes from telling BlockEntities that they are being removed without providing normal chunk unloading conditions that mod developers can expect and test for. Perhaps we should not call ``LevelChunk.clearAllBlockEntities()``, and we could instead remove them from the chunk another way without calling ``BlockEntity.setRemoved()``

fabric/src/main/java/org/valkyrienskies/mod/fabric/mixin/feature/fix_tp_ships_cross_dimension/MixinLevelChunk.java
```diff
@@ -0,0 +1,65 @@
...
+    private Map<BlockPos, LevelChunk.RebindableTickingBlockEntityWrapper> tickersInLevel;
+
+    @Shadow
+    @Final
+    private static TickingBlockEntity NULL_TICKER;
+
+    @Inject(method = "clearAllBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void clearAllBlockEntities(CallbackInfo ci) {
+        if (!this.blockEntities.isEmpty()) {
+            List<BlockEntity> blockEntitiesSnapshot = List.copyOf(this.blockEntities.values());
```
StewStrong, line 43: This could CME I think

StewStrong:
It looks pretty good to me guys :)

PR #1476, opened 2025-12-17
Commented out the look angle mixin. This time, the correct one.

zaafonin:
We all know what I'm talking about, look angle gets transformed twice when player is dragged by a ship, breaking interactions with ship (and world!) blocks. Examples include block placement, brushing suspicious sand/gravel and probably modded uses. We might need to re-test mixins into mods that compensate for this bug, Create track placement comes to mind.


zaafonin:
Track placement isn't negatively affected by this "fix", just tested it. Might be my only PR that doesn't need testing on dedicated server / prod...

jamesgreen26 (approved):
Looks good. Evy just tested this, it seems to fix the issue.

thetheaplant (approved):
does what it says on the tin

PR #1487, opened 2025-12-21
How did git break this??

blockninja124:
Git somehow nuked mine and buntings collision shape improvements, causing shafts/player heads/other blocks to have no collision


blockninja124:
This PR seems to cause VS to hang on loading a world? I'm not sure if that is my fault. Can anyone else test to confirm?

thetheaplant (approved):
does what it says on the tin

PR #1489, opened 2025-12-22
Fix Polish translations for Valkyrien Skies

KAW0:
``"key.valkyrienskies.ship_cruise": "Rejs"`` Word "Rejs" is incorrect here it means something like trip but not moving up. "Wznoszenie" which means "Going up" is better here.

``"item.valkyrienskies.physics_entity_creator": "Kreator Jednostki Fizycznej" "entity.valkyrienskies.vs_physics_entity": "Jednostka Fizyczna"`` "Jednostka" means unit like inch or kilogram it sometime can mean "ship" but here it makes it even more confusing. Since in Minecraft it's standard to translate "entity" to "byt" it fits here more. Notice that even here in  ``"command.valkyrienskies.get_ship.only_usable_by_entities": "/vs get-ship może być uruchomione tylko przez byty!"`` word "byt" is used as entity.

### DIFF
--- a/common/src/main/resources/assets/valkyrienskies/lang/pl_pl.json
+++ b/common/src/main/resources/assets/valkyrienskies/lang/pl_pl.json
@@ -3,7 +3,7 @@
 
   "tooltip.valkyrienskies.mass": "Masa",
   "category.valkyrienskies.driving": "Sterowanie",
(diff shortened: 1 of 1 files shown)


blockninja124 (approved):
LGTM

PR #1492, opened 2025-12-23
1.20.1/main dec 24 connectible chains duplicate

zaafonin:
## 🧾 What This PR Does
- Compatibility with Connectible Chains (Fabric and Forge versions)

## 🔍 How This Was Tested
- [ ] GameTest framework  
- [x] Singleplayer / Server  
- [x] Logs looked sane  
- [x] The vibes felt right *(sometimes you really do just have to vibe it)*

## ⚠️ Breaking Changes
- [ ] Yes (describe)  

---
duplicate of #1257 because this adds a gradle dep


zaafonin:
having bit of a WIP on Forge, no crash because of mixins but there's some thing handled better on the fabric side

zaafonin:
Actually I couldn't replicate this Forge crash, it happened once when the ship was crazy scaled (0.1 I think), otherwise this can be merged without conflicts. Forge version of the mod isn't even official, it's a port, Fabric version works robustly

PR #1505, opened 2025-12-25
Touchups for Create: SA mixins and other zaafslop

zaafonin:
fixes to my existing create stuff and addition mixin, also added a comment for the sculk mixin that fails on forge because it uses old mixin that can't actually mixin into injectors


ThePlasticPotato:
lgtm

PR #1512, opened 2025-12-27
shouldRender now doesn't reference ClientLevel

zaafonin:
No idea which mod even does that but this should fix it


ThePlasticPotato:
Lgtm!

PR #1527, opened 2025-12-28
Fix 0.5 offset on assembly

blockninja124:
this entire function is giga-sus but I made sure all the Vector3i's were converted to BlockPos then get the center before casting to a Vector3d. Seems to have fixed the 0.5 offset.

I also made the ship assembler item actually use the "new" assembly method, because then it actually reproduced the bug. The old assembly method didn't have this problem lmao


blockninja124:
Wait what triode already fixed this :facepalm:

blockninja124:
Never mind, triode did not fix this xD

PR #1536, opened 2025-12-30
Add get-joint stuff to GTPA

blockninja124:
Cache a (thread safe, read-only, copy) of the jointById and shipToJointIds maps from VsiPhysLevel in the GTPA for use on the gametick (by addons).


PriestOfFern (approved):
lgtm

PR #1546, opened 2026-01-02
More default masses for mod compat

Mogwump:
# ⚒️ Pull Request Offering

---
## 🧾 What This PR Does

Updates the default block masses to include blocks from more mods

- Changed Create redstone components to have the same values as the vanilla repeater and comparator
- Added the Pulse Timer from Create
- Removed redundant carved pumpkin mass in `natural/plants.json` that got overridden by the one in `functional/deco.json`
- Added support for: Autumnity, Cracker's Wither Storm Mod, Create Compressed, Create Connected, Create New Age, Create Vibrant Vaults, Galosphere, Hexcasting, Immersive Weathering, [Let's Do] Vinery, Mekanism Extras, Mekanism Generators, Nature's Spirit, Occultism, Oh The Biomes We've Gone, Woodworks (and thus leaf piles from Autumnity). Compatibility is currently not yet entirely complete for every mod, but that is also the case for existing mods like Create.
- Added masses for blocks in the tag `forge:storage_blocks/(raw_)material`

## 🔍 How This Was Tested
- [ ] GameTest framework  
- [x] Singleplayer / Server  
- [x] Logs looked sane  
- [x] The vibes felt right *(sometimes you really do just have to vibe it)*

## ⚠️ Breaking Changes
- [ ] Yes (describe)  

---

## 🧪 GameTest Oath
- [ ] Please include at least one `@GameTest`  
or  
- [x] PR does not require a `@GameTest`   
  - [x] True if this change is only a data/config/docs/logs change  
  - [x] Also True *if* you can make a good case on why you shouldn't have to add one

---

## 🜏 Declaration
- Tests _(if any)_ pass  
- No worlds shall be harmed in the making or merging of this PR 💀

### DIFF
--- a/common/src/main/resources/data/valkyrienskies/vs_mass/compat/autumnity.json
+++ b/common/src/main/resources/data/valkyrienskies/vs_mass/compat/autumnity.json
@@ -1,0 +1,37 @@
+[
+  {
+    "block": "autumnity:large_pumpkin_slice",
+    "mass": 800.0,
+    "friction": 0.3
+  },
+  {
+    "block": "autumnity:carved_large_pumpkin_slice",
+    "mass": 100.0,
+    "friction": 0.3
+  },
+  {
+    "block": "autumnity:large_jack_o_lantern_slice",
+    "mass": 100.0,
+    "friction": 0.3
+  },
+  {
+    "block": "autumnity:soul_jack_o_lantern",
+    "mass": 100.0,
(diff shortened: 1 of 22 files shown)


blockninja124 (approved):
Looks good to me, although some of the compressed blocks are rather heavy 💀.

Guess that makes sense though.

PR #1549, opened 2026-01-03
Update GTPA getJointById

blockninja124:
Before this PR, if people tried to use the GTPA to get the VSJoint from its VSJointId from **inside** the joint creation callback, they would get null, despite the joint having been created. 

This was because the GTPA created the joint, immediately called the consumer, then only updated the joint-id maps afterward.

To fix this, I've implemented a "queue" of joint consumer callbacks, which is added to on joint creation. _Then_ the joint by id maps are updated, then afterward the queued callbacks are called.

I'm not sure if my queue is the tidiest way to do this, especially using an `ArrayList`. If anyone has any suggestions on better implementations / data types, feel free to suggest a change.

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/util/GameToPhysicsAdapter.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/util/GameToPhysicsAdapter.kt
@@ -105,15 +105,21 @@ class GameToPhysicsAdapter {
             }
         }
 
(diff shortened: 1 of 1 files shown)


blockninja124:
Code to trigger the issue before this PR:
```kotlin
val gtpa = ValkyrienSkiesMod.getOrCreateGTPA(level.dimensionId)
gtpa.addJoint(joint) {
    i -> println(gtpa.getJointById(i)) // prints null, despite joint having been created
}
```

PR #1565, opened 2026-01-07
Add no collision property

blockninja124:
## What This PR Does 
- Feature
- Data/config update

## How This Was Tested
- [ ] GameTest framework  
- [x] Singleplayer / Server  
- [ ] Logs looked normal

## Breaking Changes
- [ ] Yes (describe)  
---

## GameTest Oath
- [ ] Please include at least one `@GameTest`  
or  
- [x] PR does not require a `@GameTest`   
  - [ ] True if this change is only a data/config/docs/logs change  
  - [x] Also True *if* you can make a good case on why you shouldn't have to add one
- -> I would have to add a gametest for the entire datapack (block properties) system, and I am not up for that

## Additional Notes
This does make a VS Core bug where blocks with no collision do not expand the ships AABB more noticeable. Affects of this include:
- Not being able to target the no collision block if looking at it and not any other block in the ship AABB
image
image

- Being rubber banded back when walking into the out-of-AABB block and only it [see video]
- Being unable to move onto the block outside the AABB when shifting [see video]
- Being unable to move when shifted on the block outside the AABB [see video]
- The out-of-AABB block not rendering if no other part of the ship is in view [see video]

https://github.com/user-attachments/assets/1337fd30-07f8-420a-b5cd-25849c35e314

(Video was taken with a datapack that made podzol no collision)

---

## Declaration
- Tests _(if any)_ pass  
- No worlds shall be harmed in the making or merging of this PR 💀

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
@@ -354,7 +354,7 @@ object MassDatapackResolver : BlockStateInfoProvider {
                         voxelShape = blockState.getCollisionShape(dummyBlockGetter, BlockPos.ZERO)
                     }
 
-                    val collisionShape: SolidBlockShape = if (voxelShapeToCollisionShapeMap.contains(voxelShape)) {
+                    var collisionShape: SolidBlockShape = if (voxelShapeToCollisionShapeMap.contains(voxelShape)) {
                         voxelShapeToCollisionShapeMap[voxelShape]!!
                     } else if (generatedCollisionShapesMap.contains(voxelShape)) {
                         if (generatedCollisionShapesMap[voxelShape] != null) {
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
```diff
@@ -368,7 +371,7 @@ object MassDatapackResolver : BlockStateInfoProvider {
                     val vsBlockStateInfo = map[BuiltInRegistries.BLOCK.getKey(blockState.block)]
 
                     // Create new solid block state
-                    val solidState = vsCore.newSolidStateBuilder()
+                    var solidState = vsCore.newSolidStateBuilder()
                         .shape(collisionShape)
```
zyxkad, line 381:
```suggestion
                        .shape(if ((vsBlockStateInfo?.noCollisionOverride ?: false)) vsCore.solidShapeUtils.generateShapeFromBoxes(emptyList()) else collisionShape) // use an empty shape if marked as no collision
```

common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
```diff
@@ -381,7 +384,12 @@ object MassDatapackResolver : BlockStateInfoProvider {
                         null
                     }
 
-                    VsiBlockState(solidState, fluidState)
+                    // If overrideNoCollision is set to true in datapack, force it to use air block state
+                    if ((vsBlockStateInfo?.noCollisionOverride ?: false)) {
+                        vsCore.blockTypes.airState
+                    } else {
+                        VsiBlockState(solidState, fluidState)
+                    }
```
zyxkad:
```suggestion
                    VsiBlockState(solidState, fluidState)
```

zyxkad (commented):
Should fix the AABB expansion issue

PR #1587, opened 2026-01-11
Different way of disabling worldgen in shipyard chunks (auto compat for terra, tfc, ponder world)

zaafonin:
Right now we disable worldgen in the shipyard by cancelling methods of chunk generators. This leads to boilerplate copied code for vanilla noise-based generator, vanilla flat generator, terra generator wrapper, tfc generator and any other worldgen mod we would need to be compatible with.

Really these methods are called from ChunkStatus of vanilla which is like the actual chunkgen pipeline, if there are any mods that modify it, they are rare. If we cancel calling them when necessary, we cancel any impl of these methods, vanilla or modded.

One thing I don't 100% understand is necessity of `getBaseColumn`, in vanilla it seems to be used only in structure generation which is disabled anyway.


zaafonin:
lowkey need to remove 17 more lines

ThePlasticPotato:
ts is probably fine

PR #1593, opened 2026-01-12
Make command param not useless

blockninja124:
Zaaf failed to see the very obvious function parameter :P so this is a quick fix.

Basically, makes the `deleteBlocks` boolean param of `/vs delete` work again, and sets the default back to not-deleting the blocks (because its laggy and rarely needed).


jamesgreen26 (approved):

thetheaplant (approved):
yeap

PR #1605, opened 2026-01-15
Fix Forge air pocket breathing

zaafonin:
surprisingly not my skissue, there are two basically identical mixins to updateFluidOnEyes but a wrapmethod from common was mogging the beta tail inject from forge so I wrapmaxxed it as well, priority param not needed (works without it)


LogicWheat:
I just triggered this issue lol

PR #1607, opened 2026-01-16
Fix Create tree fertilizer and Supplementaries not being compatible

zaafonin:
Create tree fertilizer "forces" a tree to grow without considering placement limitations. It does so by instantiating a "dream world" which is `TreesDreamWorld extends PlacementSimulationServerLevel`. VS2 really doesn't like new ServerLevels being randomly created.

A similar thing is done by Supplementaries and its cannon, but the cannon behavior is actually tested in their "VibeChecker" thing which runs on world load. This makes the incompatibility critical as the mod just doesn't work.

I introduce a NoVSLevelDuck interface that is tested by our mixin into ServerLevel's constructor. If the level implements this interface, all the VS-related stuff that makes the game crash later is just skipped. By doing mixins into these fake level classes (no need to implement a single method or assign a variable) we change the hierarchy and the ServerLevel init mixin will acknowledge that.


zaafonin:
Ruby kinda wanted a different solution with a list of classes, an annotation and things... I mean this one works too but he had a vision

ThePlasticPotato:
I mean if you update it with that I will merge it, but I'd like to just have A Solution At All for now

PR #1617, opened 2026-01-18
Optional shader for worldspace lighting on ships instead of Minecraft's fake light

zaafonin:
obv needs input, will describe later


ThePlasticPotato:
might have to reopen this sorry :sob:

PR #1641, opened 2026-01-24
Fix players still falling through Valkyrien Skies ships when teleporting to them if the ship is loaded for the first time this session.

Acuadragon100:
# ⚒️ Pull Request Offering

> _We shall not introduce more regressions  \
> Lest we face eternal digressions_

---
## 🧾 What This PR Does
This is a follow-up to #1566 to catch an edge case I missed the first time around.
Basically, players would still fall through ships when they were loaded for the first time since loading the world sometimes because apparently the shipAABB would not contain the entire ship immediately.

To solve this, I use the activeChunkSet to create a separate bounding box only used during the unloaded ship check. Unfortunately I can't think of a good way to detect the vertical height so the shipAABB is still used for that, I just add 16 blocks up and 16 blocks down as a precaution. This should also make sure things work properly when the player is on top of the ship rather than inside it.

Also now closes #1654

## 🔍 How This Was Tested
- [ ] GameTest framework  
- [x] Singleplayer / Server  
- [ ] Logs looked sane  
- [x] The vibes felt right *(sometimes you really do just have to vibe it)*

## ⚠️ Breaking Changes
- [ ] Yes (describe)  
- [x] No  (are you *really* sure?)

As this only uses a slightly larger bounding box for ships when checking if they are unloaded it should be fairly painless.

---

## 🧪 GameTest Oath
- [ ] Please include at least one `@GameTest`  
or  
- [x] PR does not require a `@GameTest`   
  - [ ] True if this change is only a data/config/docs/logs change  
  - [x] Also True *if* you can make a good case on why you shouldn't have to add one

Part of me wants to add a game test to make sure this bug never comes back, but I don't really know where I should put it.

> PRs without a new GameTest **will not be merged**  (If they are determined to be required) \
> _unless explicitly exempted by a maintainer._

---

## 🗝️ Additional Notes
One thing about this is that now it's basically iterating through all ships and creating a new bounding box for each one in the same dimension as the player whenever the player moves. I'm just unsure how to fix this in a better way without modifying vs-core.

---

## 🜏 Declaration
By submitting this PR, I affirm:  
- Code/data compiles and loads  
- Tests _(if any)_ pass  
- No worlds shall be harmed in the making or merging of this PR 💀

PR #1654, opened 2026-01-30
Fix (probably) players getting stuck in ships after respawning

zaafonin:
After a Player entity is reinstantiated, its collisions with ships are broken until the player relogs. I tracked this down to this recent change:
image
For collision purposes, a ship that is not in `vs_knownShips` of spawn_player_on_ship/MixinPlayer is now considered unloaded.

The `vs_knownShips` set is only populated from VSGamePackets and SeamlessChunksManager, first for server side and second for client side.
image

This creates a problem with player entities recreated without a fresh login, as player's `vs_knownShips` set is empty but the client does not receive ship load packets for the ships the player entity is supposed to be aware of (because the ships are already loaded). For now my fix is to repopulate the set from the list of loaded ships when the player entity is created. @Acuadragon100 is that an acceptable fix to your code or does it interfere with your logic?

Safety moments:
1) Level should not be null, earlier in Player/Entity constructor Level is used so if we got to TAIL of Player ctor we have a valid level.
2) Players might exist in some weird Levels that do not have a corresponding vscore ship world, in that case getShipObjectWorld gives us a dummy world. VSCore code examination (undocumented btw!) reveals `getLoadedShips` does not throw an exception returning an empty iterable instead.

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/spawn_player_on_ship/MixinPlayer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/spawn_player_on_ship/MixinPlayer.java
@@ -1,13 +1,19 @@
 package org.valkyrienskies.mod.mixin.feature.spawn_player_on_ship;
 
+import com.mojang.authlib.GameProfile;
 import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import it.unimi.dsi.fastutil.longs.LongSet;
+import net.minecraft.core.BlockPos;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.level.Level;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.valkyrienskies.mod.common.VSGameUtilsKt;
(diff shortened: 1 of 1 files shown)


Acuadragon100:
Ah... Right, that variable will be reset when the player respawns...
Didn't think about that...

Only thing that worries me slightly is that your fix repopulates the vs_knownShips variable on the server-side, which could contain ships the client hasn't loaded yet. If I would have written this I'd probably only repopulate it client-side, then either send that to the server via PacketChangeKnownShips or just copy the old set in ServerPlayer::restoreFrom.

With that said, I did some brief tests and it doesn't seem like your fix breaks anything. I did notice the player still falls through ships (specifically when another player has already loaded the ship), but that also seems to happen without your commit so that's probably my fault...

TLDR; You can probably merge this and I'll just fix it later if necessary.

Acuadragon100:
Actually wait...
I think my dev environment is playing tricks on me, because I did a clean build without your commit and suddenly my fix for players falling through ships (#1641) seemed to work fine again...

Acuadragon100:
Oh well, it's up to you if you want to try to make this work alongside #1641 (in that case you might want to pull request it into that branch, or just let me know I have permission to cherry-pick your commit into my branch) or if you want to merge this as-is and let me fix it later. Will GitHub automatically notify me when this is merged? Either way, feel free to ping me when it's merged so I can reopen #1641.

zaafonin:
I don't mind it being cherry picked at all, it is your feature so you know better what's happening with it

Acuadragon100:
Ah, neat. In that case I'll just merge this into #1641 and reopen it.

Acuadragon100:
Ok, I've now merged this into #1641.
Thanks for letting me know about this bug!

blockninja124:
Superseded by https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1641

PR #1658, opened 2026-02-01
Disable client connectivity by default

blockninja124:
Goes along with https://github.com/ValkyrienSkies/vs-core/pull/98

It's unfortunate this will kill air pockets for the time being, but conn is just too laggy (because of bugs) to justify it being on by default 😔

Hopefully the Valkyrien Air stuff better air pockets can be finished for a (soonish) future release.


blockninja124:
Now a draft because it’s waiting on a new VS core version to disable server conn

PR #1672, opened 2026-02-07
Fix util function

blockninja124:
## What this PR does:
- [x] Bug fix 
- [ ] Feature
- [ ] Compat with another mod

This PR fixes a bug introduced in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1368 which made the VSGameUtils function `transformFromWorldToNearbyShipsAndWorld` not behave as it's KDoc suggested. 

While the KDoc said that it "calls the [cb] with the AABB itself", the water PR removed this functionality. I don't know why the water mixin needed it to not call with the original AABB, but I don't want to go down that rabbit hole. So instead, I added a new util function `transformFromWorldToNearbyShips`, which does not call the `cb` with the original aabb. This new function is used in the MixinEntities, and the old function repaired to its stated functionality.

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [x] Yes (describe)
- [ ] No

Technically they could be breaking changes if addons relied on this function _not_ calling the cb... but since it was a bug not intended functionality (as made clear by the KDoc), that's for the addons to figure out (if any).

---

## Additional Notes

Not sure flowing water on ships is still working 100%, even before this PR. It's all a little sus tbh. But a full rewrite is for... later.

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/VSGameUtils.kt
@@ -205,6 +205,20 @@ private fun getShipObjectManagingPosImpl(world: Level?, chunkX: Int, chunkZ: Int
  */
 fun Level.transformFromWorldToNearbyShipsAndWorld(aabb: AABB, cb: Consumer<AABB>) {
     val tmpAABB = AABBd()
+    cb.accept(aabb)
+    getShipsIntersecting(aabb).forEach { ship ->
+        cb.accept(tmpAABB.set(aabb).transform(ship.worldToShip).toMinecraft())
+    }
+}
+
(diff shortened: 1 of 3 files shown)


cylian914:
looks good
again sorry for breaking things

blockninja124:
And I'm sorry for approving the PR without spotting it the first time :facepalm: 
I'll be better next time :pray: :pleading_face:

alex-s168 (approved):

jamesgreen26 (approved):

Issue #1702, opened 2026-02-26
Client is stuck in an infinte loop in the render thread when the player teleports large distances while a ship is loaded.


NythicalNorm:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods

- [x] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

Teleporting more than 10,000,000 blocks hangs the render thread, it will infinitely execute this for loop in Frustrum class in net.minecraft.client.renderer.culling

<img width="1020" height="327" alt="Image" src="https://github.com/user-attachments/assets/dfb2d668-5308-4dbc-a4b7-fc5a94d5d8bc" />

this.camX and this.camY will get out of bounds and approach infinity.

I have tested this and it happens on VS 2.4.6 on forge and 2.4.10 on both forge and fabric. (also occurs on the current master VS branch from a couple days ago)

btw this also happens if you teleport to a different dimension and your position in that dimensions is also like 10,000,000 plus that is how i found this issue.

if you have Embeddium installed this issue doesn't occur, i assume the same is true of sodium on fabric. anyway i could do a naive mixin into Frustum to fix this for my mod but this is an actual issue with VS,

### Issue reproduction

its easy to recreate, load up a world make sure a ship is visible on the client and next to the player and do "/tp @a 10000000 200 0"

<img width="1920" height="1009" alt="Image" src="https://github.com/user-attachments/assets/be992dc2-cdb1-43fb-81d8-2515240652c6" />

### Logs

This is what the log output was before the hang:

[tp_hang_log.txt](https://github.com/user-attachments/files/25578395/tp_hang_log.txt)

blockninja124:
If you have a mixin that could fix it, why not just make a PR to fix the code directly?

NythicalNorm:
I don't have a mixin but i think i can make something from what it looks like in the debug, if i can fix it I'll make a PR.

blockninja124:
Fixed by https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1703

PR #1703, opened 2026-02-26
Fix for a infinite loop in the Frustum class on the Client

NythicalNorm:
## What this PR does:
- [x] Bug fix 
- [ ] Feature
- [ ] Compat with another mod

When a player teleports millions of blocks in coordinates away from where they are while they have a VS ship loaded on the client, the projection matrix or something gets garbled causing the client to hang due to a infinite loop in the offsetToFullyIncludeCameraCube method in Frustum class in net.minecraft.client.renderer.culling, I made a issue for this #1702 ,
but here is quick fix. <br>
Also for reference [here](https://github.com/iPortalTeam/ImmersivePortalsMod/blob/1.20.1/imm_ptl_core/src/main/java/qouteall/imm_ptl/core/mixin/client/render/MixinFrustum_FixDeadLoop.java) is the same issue being fixed in Immersive Portals I could not find the root cause for the garbled projection matrix (or whatever it may be, the comments on Immersive Portals fix said it was the projection matrix) after some looking, but even if the root cause is fixed I think this is good to have just in case. though I didn't copy the code (still immersive portals is apache 2.0).

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [x] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [x] No

---

## Additional Notes
Tested that its Fixed in both Forge and Fabric, Sodium and Embeddium seem to fix this issue on their own, but i built a jar and tested that this fix is compatible with Sodium and Embeddium. Also from the Immersive Portals fix of this issue, I am pretty sure this mixin works as is for 1.21.1.

There does seem to some server side weirdness that I noticed that can be reproduced by going to the world border, making a ship there, building a nether portal and entering the nether while the ship is loaded, this causes the chunk generation threads to go all weird, they eventually load the nether but there does seem to be a major slow down. but this is unrelated to this fix and it could be caused by other mods in the VS dev environment.

---

## Declaration
- Tests _(if any)_ pass  
Client side fix so no worries here 😅


NythicalNorm:
lol, accidentally closed it by syncing.

blockninja124 (dismissed):
LGTM

PR #1710, opened 2026-03-04
Make liquidMaterialToDensityMap public

Gotve1:
## What this PR does:
- [ ] Bug fix 
- [ ] Feature
- [+] Code refactor / improvement
- [ ] Compat with another mod

makes liquidMaterialToDensityMap public until someone fix MassDatapackResolver

---

- [+] Forge
- [+] Fabric

## Where this PR was tested:
- [+] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [+] No

---

## Additional Notes

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
@@ -277,6 +277,15 @@ object MassDatapackResolver : BlockStateInfoProvider {
      *
      * It is **not recommended** to call this yourself!
      */
+
+    val liquidMaterialToDensityMap: HashMap<Fluid, Pair<Double, Double>> = hashMapOf(
+        Fluids.WATER to Pair(1000.0, 0.3),
+        Fluids.LAVA to Pair(10000.0, 1.0),
+        Fluids.FLOWING_WATER to Pair(1000.0, 0.3),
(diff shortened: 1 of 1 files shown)


blockninja124:
Wouldn't it be better to just... identify and fix the issue with the datapack resolver?

common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
```diff
@@ -277,6 +277,15 @@ object MassDatapackResolver : BlockStateInfoProvider {
      *
      * It is **not recommended** to call this yourself!
      */
+
+    val liquidMaterialToDensityMap: HashMap<Fluid, Pair<Double, Double>> = hashMapOf(
```
blockninja124, line 281: I don't think this is even a mutable map, so there isn't any point to exposing it? Addons will just be able to read it, not write to it.

zyxkad: HashMap is mutable

blockninja124: In kotlin? You usually have to use `mutableMapOf()` instead of `mapOf()`, but I don't know if that applies to hash maps.

PR #1712, opened 2026-03-06
Fix: Y velocity accumulates during horizontal collision

### LINKED ISSUE #1681: Speed is applied wrongly when walking into sides of blocks

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

Whenever walking into a surface, speed is applied downwards. in Vanilla, walking into a block and removing it underneath would start you at 0 vertical speed before accelerating. However, when valkyrian skies and Clockwork are installed, walking into a wall while breaking the block underneath the player results in the player being thrown downwards.

### Issue reproduction

1. Load Minecraft Forge 1.20.1 with valkyrian skies, clockwork, and its dependencies installed.
2. load up a fresh world and build a high tower.
3. place a block to stand on at the sides of the tower.
4. walk into the sides of said tower while breaking the block underneath the player.
5. the player should be thrown downwards with a lot of speed.

### Logs

[debug.log](https://github.com/user-attachments/files/25289964/debug.log)
[latest.log](https://github.com/user-attachments/files/25289965/latest.log)

clawdbot-silly-waddle:
`redirectSetVelocity` in MixinEntity uses `@Inject + callbackInfo.cancel()` to replace vanilla's axis-zeroing with projection-based velocity removal. But `cancel()` on an `@Inject` at an INVOKE target cancels the *entire remaining method*, not just the targeted call. This skips everything after `setDeltaMovement(DDD)V` in `Entity.move()`, including `block.onEntityLand()` which resets Y velocity on ground contact.

Without `onEntityLand`, Y velocity accumulates from gravity every tick while grounded (~-0.08/tick). After 100 ticks against a wall: vel.y ≈ -3.4. On a bed, `BedBlock.onEntityLand` bounces the player (`vel.y = -vel.y * 0.66`), converting this into a ~20 block upward launch. No sneaking required - just walking into a wall while standing on a bed is enough.

Found via https://www.reddit.com/r/feedthebeast/comments/1rlk2dm/weird_launch_bug_that_only_happens_in_my_fabric/

This affects any entity with horizontal collision (wall contact), not just beds - beds just make it visible because of the bounce. On regular blocks the accumulated velocity is silently zeroed when horizontal collision stops. Other skipped behavior includes `block.onSteppedOn()`, the velocity multiplier (soul sand/honey), and fire checks.

The bug was technically introduced in [541f58d5](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/541f58d5925a5a520a9d1d879265e9cf3d48eae6) (2021-03-23) when the mixin changed from `@Redirect` to `@Inject + cancel`, but was masked because the Y velocity was set from `movementAdjustedForCollisions.y` (which is 0 during ground collision). It became visible after [fdb33aa0](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/fdb33aa0a31f1c9692465efaf834b2a700b62279) ("THE GREAT UNFUCKENING PART 1", 2025-06-22) which correctly changed to using `getDeltaMovement().y` for the projection - but this meant the accumulated Y velocity was no longer being overwritten.

The fix replaces `@Inject + cancel` with `@WrapOperation`, which wraps only the `setDeltaMovement(DDD)V` call and lets all subsequent vanilla code run normally. The projection math is unchanged. The mixin is in `common/` so this affects both Fabric and Forge.

This was partially caught before in #955 where @MoePus noticed `tryCheckInsideBlocks` was being skipped. @ewoudje commented:

> Yep nice catch, Im kinda worried if we the mixin also skips other things such as checking for fire, or popping the profiler.

This turned out to be right - `onEntityLand` and several other calls were also missing.

Tested in-dev singleplayer on both Fabric 1.20.1 and Forge 1.20.1. Positive testing confirms the issue is fixed on both loaders. Negative testing (checking for regressions) was limited to basic gameplay - more thorough testing of player collision scenarios would be appreciated.

Related commits:
- [b1aec085](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/b1aec085f04e981be74d61cdd86c8fe48445907d) (2021-03-21) - replaced `@Overwrite` of `move()` with `@Redirect` (which correctly preserved `onEntityLand`)
- [541f58d5](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/541f58d5925a5a520a9d1d879265e9cf3d48eae6) (2021-03-23) - changed to `@Inject + cancel`, accidentally skipping `onEntityLand`
- [fdb33aa0](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/fdb33aa0a31f1c9692465efaf834b2a700b62279) (2025-06-22) - changed Y velocity source from `movementAdjustedForCollisions` to `getDeltaMovement()`, unmasking the bug
- [be136874](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/be13687497e80201fb1d4c13198a3f005eddb3ee) / #955 (2024-09-23) - restored `tryCheckInsideBlocks` but not `onEntityLand`

### Test demo

Before fix (velocity accumulates, bed bounce launches player ~20 blocks):

https://cb.2d.rocks/files/github-assets/vs-fix-bed-launch-unfixed.mp4

After fix (velocity stays stable, player stays on bed):

https://cb.2d.rocks/files/github-assets/vs-fix-bed-launch-fixed.mp4

Fixes #1681

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
@@ -25,7 +25,7 @@ import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
-import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
 import org.valkyrienskies.core.api.ships.Ship;
 import org.valkyrienskies.mod.common.VSGameUtilsKt;
 import org.valkyrienskies.mod.common.util.EntityDraggingInformation;
@@ -113,28 +113,23 @@ public abstract class MixinEntity implements IEntityDraggingInformationProvider 
     }
 
     /**
-     * This mixin replaces the following code in Entity.move().
-     *
-     * <p>if (movement.x != vec3d.x) { this.setVelocity(0.0D, vec3d2.y, vec3d2.z); } </p>
-     *
-     * <p>if (movement.z != vec3d.z) { this.setVelocity(vec3d2.x, vec3d2.y, 0.0D); } </p>
-     *
-     * <p>This code makes accurate collision with non axis-aligned surfaces impossible, so this mixin replaces it. </p>
+     * Replaces vanilla's axis-zeroing collision response with projection-based velocity removal,
+     * enabling accurate collision with non-axis-aligned ship surfaces.
      */
-    @Inject(method = "move", at = @At(
+    @WrapOperation(method = "move", at = @At(
         value = "INVOKE",
-        target = "Lnet/minecraft/world/entity/Entity;setDeltaMovement(DDD)V"),
-        locals = LocalCapture.CAPTURE_FAILHARD, cancellable = true
-    )
-    private void redirectSetVelocity(final MoverType moverType, final Vec3 movement, final CallbackInfo callbackInfo,
-        final Vec3 movementAdjustedForCollisions) {
+        target = "Lnet/minecraft/world/entity/Entity;setDeltaMovement(DDD)V"))
+    private void redirectSetVelocity(final Entity instance, final double x, final double y, final double z,
+        final Operation<Void> original,
+        @Local(argsOnly = true) final Vec3 movement,
+        @Local(ordinal = 1) final Vec3 movementAdjustedForCollisions) {
 
         // Compute the collision response horizontal
         final Vector3dc collisionResponseHorizontal =
             new Vector3d(movementAdjustedForCollisions.x - movement.x, 0.0,
                 movementAdjustedForCollisions.z - movement.z);
 
-        // Remove the component of [movementAdjustedForCollisions] that is parallel to [collisionResponseHorizontal]
+        // Remove the component of velocity that is parallel to the collision normal
         if (collisionResponseHorizontal.lengthSquared() > 1e-6) {
             final Vec3 deltaMovement = getDeltaMovement();
 
@@ -151,10 +146,6 @@ public abstract class MixinEntity implements IEntityDraggingInformationProvider 
                     - collisionResponseHorizontalNormal.z() * parallelHorizontalVelocityComponent
             );
         }
-        // The rest of the move function (including tryCheckInsideBlocks) is skipped, so calling it here
-        tryCheckInsideBlocks();
-        // Cancel the original invocation of Entity.setVelocity(DDD)V to remove vanilla behavior
-        callbackInfo.cancel();
     }
 
     // endregion
@@ -294,9 +285,6 @@ public abstract class MixinEntity implements IEntityDraggingInformationProvider 
 
     @Shadow
     public abstract void setDeltaMovement(double x, double y, double z);
-
-    @Shadow
-    protected abstract void tryCheckInsideBlocks();
 
     @Shadow
     protected abstract Vec3 collide(Vec3 vec3d);


clawdbot-silly-waddle:
@blockninja124

> Its Valkyrien Skies. We haven't figured out why it causes this yet unfortunately. You say it doesn't happen on forge? That's interesting, unless you just don't have VS on forge.

fyi this fixes it :D

clawdbot-silly-waddle:
This should also fix #1681 — same root cause (Y velocity accumulating while walking into a wall).

blockninja124 (approved):
I'm not sure I approve of bot-written PRs, but I've scrutinized this PR to the best of my ability and it does what it claims, and in probably the best way. 

My only complaint is the javadoc on the mixin has gotten more vague and confusing, but it was also vague and confusing before, so whatever I guess.

zaafonin:
@clawdbot-silly-waddle , you have written GameTests to perform the test as shown in the videos. Is there a chance you’ll include them into the PR as well? They would be very valuable as VS2 devs want to implement more GameTests in the future.

clawdbot-silly-waddle:
The test in the videos isn't a GameTest - it's a standalone Fabric mod that automates the repro steps. Fabric API does have a client GameTest API that supports input simulation (as would be required here to walk against a wall), but it's still marked experimental. Either way, VS doesn't currently use client gametests and this fix is in `common/` (cross-loader).

StewStrong (approved):
Looks very nice, thanks for the fix! For future reference it would be nice to have gametests in the future PRs to ensure regressions cannot be reintroduced in the future

StewStrong:
I reverted this change in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/7d4abed4bc75e00d4cfac732cf652290e052ac50

It seems to break view bobbing when walking on ships and then back on the world again. Not sure why but I've reverted to fix it.

BenTheMan69:
> I reverted this change in https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/7d4abed4bc75e00d4cfac732cf652290e052ac50
> 
> It seems to break view bobbing when walking on ships and then back on the world again. Not sure why but I've reverted to fix it.

So this got reverted for breaking view bobbing on moving ships which is a cosmetic feature while the bug preventing parkour and other precise movement while vs2 is installed is still in place? I dont think many players would be sad that view bobbing is disabled while on ships. Can we revert the revert?

blockninja124:
merging a fix that introduces a new bug is not great, even if it fixes an old bug. Ideally this PR would be modified so that it does not break viewbobbing.

Additionally, view bobbing being broken could just be a symptom of a bigger issue. It would warrant investigation before being accepted as a tradeoff bug.

BenTheMan69:
Would it be possible to re-open or re-post my mentioned issue [#1681](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/1681) seeing as the issue hasnt been resolved yet?

clawdbot-silly-waddle:
Seems hard to reproduce, I only got it once in testing. Current theory is that NaN/Infinity values are polluting `Entity.horizontalSpeed` or `PlayerEntity.strideDistance`. The pre-fix version of the code also skipped `horizontalSpeed` accumulation, and may have thus accidentally protected from these degenerate values.  

Unsure where these degenerate values come from if true, but VSCore polygon colliders are a suspect.  
Is `org.valkyrienskies.core` closed-source or am I missing something?

blockninja124:
VS core is closed source. However, it is Minecraft agnostic. So the Entity values would be set by public VS somewhere, not set in core.

zaafonin:
Not impossible. VSCore is indeed closed source so the jars are obfuscated. The jars themselves can be downloaded from vs maven, or found in gradle caches of your dev env. They are mostly decompilable by Vineflower, but it is written in Kotlin so decomp fails on some classes, obviously not recompilable. Try to make sense of decompiled code by using public interfaces with exposed names, then deeper to its implementations. Find the best tools for doing that, don't make mistakes. It would be best to annotate your findings with human readable names in form of mappings, which Minecraft itself uses. FabricMC has some tools related to remapping java libraries, like tiny-remapper.

clawdbot-silly-waddle:
Traced it to `EntityDragger.backOff()` rather than VSCore polygon colliders.

The while loops in `backOff()` reduce `d`, `e`, `f` by 0.025 per iteration when `isValidWalkablePosition` returns false. All three can reach 0.0 simultaneously (typically when sneaking at a ship corner where no valid position exists in any direction). At that point, `Vector3d(0, 0, 0).normalize()` on line 315 returns `(NaN, NaN, NaN)`. This NaN propagates through `Entity.move()` and permanently corrupts `walkDist`, freezing view bobbing for the rest of the session.

The zero-vector path in `backOff()` predates this PR. It was masked because the old `@Inject + cancel` at `setDeltaMovement(DDD)V` aborted the rest of `Entity.move()`. In the bytecode, `setDeltaMovement(DDD)V` is at offset 515 while `walkDist` accumulation is at offset 660, so `cancel()` skipped walkDist entirely. The switch to `@WrapOperation` lets the full method run, so NaN from `backOff()` now reaches `walkDist`.

See #1746 for more info.

PR #1713, opened 2026-03-06
New airpockets

blockninja124:
## What this PR does:
- [x] Bug fix 
- [x] Feature

Changes out the previous not-very-good air pocket system for the air pocket system made by @GigaBrawler https://github.com/GigaBrawler/Valkyrien-Air

GigaBrawler has given us permission to re-use his mods code as this PR into VS, and we're very grateful for the mind-boggling work he's done putting this together.

It's not quite perfect (I explain more in the additional notes), but it's pretty close to a state ready for release. Once a few final fixes on this are done, I think it will be time for VS 2.4.11. 

Soon:tm:

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Additional Notes

Remaining problems:
- Flooding is a bit iffy when the hole is right at the water surface
- Flooding is too fast. This is not fixable by some magic value, it's because it fills water blocks based on a plane, not block-by-block. We could maybe just slow the plane down a lot, but I think it would be better to try and come up with some block-by-block flooding spreading out from the hole.
- Some remaining visual issues, including
   - Visual artifacts on some AMD graphics cards(?)
   - No water fog or water texture when looking at the side of water below the surface
   - Water surface is visible if shaders are on, or physics mod pro waves are used
   - Occasional air-pocket or water-surface artifacts when moving camera at specific angles

However none of these visual issues I would consider a release blocker, they can all be improved / fixed in a proceeding update.

---

## Declaration
- Tests _(if any)_ pass  


blockninja124:
To clarify: I think this PR will be ready to review once we figure out how to slow flooding down. Right now its pretty close to instantaneous, hole becomes layer filled with water. There are some values for flooding speed that seem quite slow, but maybe we can tweak them to be even slower.

PR #1714, opened 2026-03-06
Remove "Collision box is too big" log

blockninja124:
## What this PR does:
idk what category this is ngl

Removes the "Collision box is too big! Returning empty list, this may break things" log. This log triggers way too much with some other mods installed, and causes people to think there is an issue when in reality there isn't. It has never been helpful in debugging an issue anyway

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/get_entities/MixinLevel.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/get_entities/MixinLevel.java
@@ -56,9 +56,9 @@ public class MixinLevel {
         final Predicate<? super T> predicate, final CallbackInfoReturnable<List<T>> cir) {
 
         if (BugFixUtil.INSTANCE.isCollisionBoxTooBig(area)) {
-            LIMITER.maybeRun(() ->
+            /*LIMITER.maybeRun(() ->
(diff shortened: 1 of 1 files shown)


PGgamer2:
When this message pops up sometimes things actually break. As an example, I'm using VS2 with cosmic horizons and starlance and whenever i change dimension the error is printed in the console and the ship disappears until i log back onto the world.
I've removed the AABB size limit and it doesn't break things anymore (but the game takes a massive frame drop).
I'd suggest to instead leave the error message as-is and resize the AABB to the max size to minimize damage

blockninja124:
A maximum size AABB would cause exactly the same problems as an empty one.

PGgamer2:
By "maximum size" I mean one set arbitrarily. BugFixUtil calculates for a max area of 100.000.000, so it could resize the AABB to comply to this requirement

blockninja124:
Ok but that would still cause problems. Because the original AABB is stretching from the shipyard to the world, any clamped AABB is not going to reach anywhere near the original area expected.

PR #1728, opened 2026-03-15
Update changelog and change version

blockninja124:

Bumps the version to 2.4.11 and adds a changelog.

After an hour long argument in VS creators, I'm defeated. I want a new version scheme `a.b.c.d` so we can use `c` for medium updates and `d` for bug fixes, but everyone else despises having an actually useful semver so whatever. The useless `2.4.11` it is.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,8 @@
-## VS 2.5.0.0
+## VS 2.5.0
 
-Proper air pockets! As well as a new version scheme, and many small fixes.
+Proper air pockets! As well as a many small fixes.
+
+**Note: PhysX is still not the default. We have just chosen to use 2.5 for the air pocket release instead.**
 
 #### Changes:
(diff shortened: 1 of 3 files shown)


zaafonin:
my point is adding water occlusion not just for complete air pockets but bowl/boat shaped (and even flooding/draining physics so it's not just visual) is not even a medium update it's a major one; the problem is trying to fit major updates into the regular release cycle of bug fixes (while the latest bug fix update is kinda broken and the fixes wait on the major thing)

tl;dr version for tiktokkers: this is 2.5 worthy, 2.4.11 should have released without waiting on all that, versioning isn't the problem the process is

(yea said that in discord but just in case someone didn't read through all of that)

blockninja124:
I don't care if its 2.4 or 2.5 worthy, I want my double 0 at the end, which you have a problem with.

PR #1730, opened 2026-03-15
Ship weather occlusion

copperwarrior:
## What this PR does:
- [X] Bug fix 
- [X] Feature
- [ ] Code refactor / improvement
- [ ] API addition / improvement
- [ ] Compat with another mod

**Explain what this PR is doing:**
Adds a method in CompatUtils.kt to calculate heightmap at position inclusive with ships via adding a worldheight-down clip to existing heightmap.
Uses this method to mixin to rain and snow renderer and rain tick so that rain is occluded by VS ships.

---

## Modloaders this PR affects:
- [X] Forge
- [X] Fabric

## Where this PR was tested:
- [X] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [ ] No

---

## Additional Notes

Slight performance concerns when the CompatUtils method is searching for BLOCKING_NO_LEAVES, in order to maintain parity leaf blocks have to be manually skipped by looping consecutive clips

Returned voxelshapes aren't rotated correctly, any solution to this would be well out of scope of this feature

Rain splash particles, as any particle, does not collide with the ship so in the rain roofs can be a bit "leaky". Fix for this would be out of scope for this feature.

---

## Declaration
By submitting this PR, I affirm:  
- Code/data compiles and loads  
- Tests _(if any)_ pass  
- This PR will not brick worlds

PR #1731, opened 2026-03-15
Particles that collide with blocks now consider ships

copperwarrior:
## What this PR does:
- [X] Bug fix 
- [X] Feature
- [ ] Compat with another mod

Redirects particle usage of `Entity.collideBoundingBox` so that particles that should collide with blocks properly collide with ships.

https://github.com/user-attachments/assets/3490dcc3-3f13-4de8-8682-4175b399c1f3

---

- [X] Forge
- [X] Fabric

## Where this PR was tested:
- [X] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [X] No

---

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/util/EntityShipCollisionUtils.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/util/EntityShipCollisionUtils.kt
@@ -27,6 +27,7 @@ import org.valkyrienskies.mod.util.BugFixUtil
 import java.util.stream.Stream
 
 object EntityShipCollisionUtils {
+    private const val PARTICLE_COLLISION_BOX_EXPANSION = 0.00390625 //1.0 / 256.0
 
     private val collider = vsCore.entityPolygonCollider
 
@@ -123,8 +124,14 @@ object EntityShipCollisionUtils {
             return movement
         }
 
+        val collisionBoundingBox = if (entity == null) {
+            entityBoundingBox.inflate(PARTICLE_COLLISION_BOX_EXPANSION)
+        } else {
+            entityBoundingBox
+        }
+
         val (newMovement, shipCollidingWith) = collider.adjustEntityMovementForPolygonCollisions(
-            movement.toJOML(), entityBoundingBox.toJOML(), stepHeight, collidingShipPolygons
+            movement.toJOML(), collisionBoundingBox.toJOML(), stepHeight, collidingShipPolygons
         )
         if (entity != null) {
             val standingOnShip = entity.level().getLoadedShipManagingPos(entity.onPos)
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/util/EntityShipCollisionUtils.kt
```diff
@@ -27,6 +27,7 @@ import org.valkyrienskies.mod.util.BugFixUtil
 import java.util.stream.Stream
 
 object EntityShipCollisionUtils {
+    private const val PARTICLE_COLLISION_BOX_EXPANSION = 0.00390625 //1.0 / 256.0
```
zyxkad, line 30: why filling the pre-calculate value instead of just write the expression out? I believe compiler are smart enough to do the calculation at compile time.

copperwarrior: I'm used to pre-calculating expressions like this from other projects. If this is an issue, I can change this to the expression.

common/src/main/kotlin/org/valkyrienskies/mod/common/util/EntityShipCollisionUtils.kt
```diff
@@ -123,8 +124,14 @@ object EntityShipCollisionUtils {
             return movement
         }
 
+        val collisionBoundingBox = if (entity == null) {
+            entityBoundingBox.inflate(PARTICLE_COLLISION_BOX_EXPANSION)
```
blockninja124, line 128: Can you maybe explain why we do this? Cause I for one have no idea why we inflate the particle collision box by this magic value lol (I'm sure its some mojank or something, I just think a short explanation would be handy)

blockninja124 (approved):
Tiny suggestion but otherwise LGTM

common/src/main/kotlin/org/valkyrienskies/mod/common/util/EntityShipCollisionUtils.kt
copperwarrior: I guess I forgot to port my comment from my testing branch.
Basically, I was doing testing with #1730 and rain splash particles were spawning on the very edges of blocks and falling through. Inflating the bounding box very slightly resolved this issue. I don't personally see a reason why this would cause any noticeable issues.

jamesgreen26:
Screenshot_2026-03-17_at_6 28 53_PM

Works

jamesgreen26 (approved):

PR #1746, opened 2026-03-22
Fix Y velocity accumulation during horizontal collision + NaN in backOff()

### LINKED ISSUE #1681: Speed is applied wrongly when walking into sides of blocks

clawdbot-silly-waddle:
Resubmission of #1712 with an additional fix for the NaN bug that caused the revert (`7d4abed4`).

**Fix 1: Y velocity accumulation (from #1712)**

`redirectSetVelocity` in MixinEntity used `@Inject + cancel` to replace vanilla's axis-zeroing with projection-based velocity removal. `cancel()` aborted the rest of `Entity.move()` after the `setDeltaMovement(DDD)V` call, skipping `block.updateEntityAfterFallOn` (which resets Y velocity on ground contact), `walkDist` accumulation, `tryCheckInsideBlocks`, and other post-collision behavior. Without `updateEntityAfterFallOn`, Y velocity accumulated from gravity every tick while grounded. On beds, this converted into upward launches via `BedBlock.bounceUp`.

The fix replaces `@Inject + cancel` with `@WrapOperation`, wrapping only the `setDeltaMovement(DDD)V` call and letting the rest of `Entity.move()` run. Full details in #1712.

**Fix 2: NaN from zero-vector normalize in backOff()**

`EntityDragger.backOff()` transforms player movement into ship space and uses while loops to reduce the components (`d`, `e`, `f`) until `isValidWalkablePosition` returns true for each direction. When no valid position exists in any direction (typically at a ship corner while sneaking), all three reach 0.0. `Vector3d(0, 0, 0).normalize()` returns `(NaN, NaN, NaN)`, which propagates through `Entity.move()` and permanently corrupts `walkDist`, freezing view bobbing for the rest of the session.

This was masked by the old `@Inject + cancel` because `cancel()` aborted `Entity.move()` before reaching `walkDist` accumulation (bytecode offsets 515 vs 660 in `Entity.move`). Fix 1's `@WrapOperation` lets the full method run, exposing the NaN.

The fix adds a zero-vector guard after the while loops: if `d == 0.0 && e == 0.0 && f == 0.0`, return `Vec3(0.0, vec3.y, 0.0)` instead of calling `normalize()`. This preserves the gravity component while preventing the NaN.

**Known limitations of backOff()**

`backOff()` is still fundamentally flawed. The NaN guard is a symptom fix for the specific issue reported (frozen view bobbing). Known remaining issues:
- Players can't crouch to their hitbox edge on ships like vanilla allows
- Sneaking while falling onto a rotated ship shifts the player sideways
- Several architectural deviations from vanilla's `maybeBackOffFromEdge`: single-point raycasts instead of bounding box collision checks, reduces all 3 axes (d/e/f) instead of just horizontal (d/f), coordinate space mixing in the `level.clip` raycast

Tested on Fabric 1.20.1 in singleplayer. View bobbing freeze no longer occurs.

Fixes #1681

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
@@ -25,7 +25,7 @@ import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
-import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
 import org.valkyrienskies.core.api.ships.Ship;
 import org.valkyrienskies.mod.common.VSGameUtilsKt;
 import org.valkyrienskies.mod.common.util.EntityDraggingInformation;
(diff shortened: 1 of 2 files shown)


clawdbot-silly-waddle:
This also fixes briefly getting stuck in the air when sneaking while falling onto an angled ship. Same root cause: `backOff()` returns NaN, which corrupts the movement vector and causes the entity to freeze mid-air.

PR #1751, opened 2026-03-25
deltaMovement Y velocity buildup patch

hehdawn:
## What this PR does:
- [x] Bug fix 

Prevents deltaMovement Y axis buildup when moving into a solid object.

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer

## Breaking Changes
- [x] No

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
@@ -174,7 +174,7 @@ public abstract class MixinEntity implements IEntityDraggingInformationProvider 
             target = "Lnet/minecraft/world/entity/Entity;setDeltaMovement(DDD)V"
         )
     )
-    private void CancelsetDeltaMovement(Entity instance, double d, double e, double f) {}
+    private void cancelSetDeltaMovement(Entity instance, double d, double e, double f) {}
 
 
     // region Block standing on friction and sprinting particles mixins
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
@@ -4,6 +4,7 @@ import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
 import net.minecraft.core.BlockPos;
+import net.minecraft.network.chat.Component;
 import net.minecraft.util.RandomSource;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityDimensions;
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
```diff
@@ -136,29 +150,33 @@ private void redirectSetVelocity(final MoverType moverType, final Vec3 movement,
...
 
+    // this cancels the actual setDeltaMovement. We can't cancel it in the @Inject, that cancels the entire move function
+    @Redirect(
+        method = "move",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/world/entity/Entity;setDeltaMovement(DDD)V"
+        )
+    )
+    private void CancelsetDeltaMovement(Entity instance, double d, double e, double f) {}
```
blockninja124:
```suggestion
    private void cancelSetDeltaMovement(Entity instance, double d, double e, double f) {}
```
nicer name

common/src/main/java/org/valkyrienskies/mod/mixin/feature/entity_collision/MixinEntity.java
```diff
@@ -136,29 +150,33 @@ private void redirectSetVelocity(final MoverType moverType, final Vec3 movement,
...
         }
-        // The rest of the move function (including tryCheckInsideBlocks) is skipped, so calling it here
-        tryCheckInsideBlocks();
-        // Cancel the original invocation of Entity.setVelocity(DDD)V to remove vanilla behavior
-        callbackInfo.cancel();
     }
-
     // endregion
 
+    // this cancels the actual setDeltaMovement. We can't cancel it in the @Inject, that cancels the entire move function
```
blockninja124:
```suggestion
    // This cancels the actual setDeltaMovement. We can't cancel it in the @Inject, that cancels the entire move function. Somehow this doesn't interfere with the @Inject, crazy.
```

blockninja124 (dismissed):
couple of minor tweaks but otherwise looks good. 

Fixes the same thing as https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1746, but seems to be a smaller fix.

Not sure which PR is the better one, but I have seen this one in action and it passes all the general tests for movement.

blockninja124 (approved):

clawdbot-silly-waddle:
This has the same view bobbing bug that caused #1712 to be reverted (https://github.com/ValkyrienSkies/Valkyrien-Skies-2/commit/7d4abed4bc75e00d4cfac732cf652290e052ac50).

Repro: sneak while falling onto an angled ship. `EntityDragger.backOff()` while loops reduce all three components to 0.0, then `Vector3d(0, 0, 0).normalize()` returns `(NaN, NaN, NaN)`. The NaN propagates through `Entity.move()` and permanently corrupts `walkDist` (= `horizontalSpeed` in Yarn), freezing view bobbing for the rest of the session.

The old `@Inject + cancel` masked this because `cancel()` aborted `Entity.move()` before reaching the `walkDist += vec3d.horizontalLength()` accumulation (bytecode offset 515 vs 660). Both this PR and #1712 let `Entity.move()` continue past that point, so the NaN reaches `walkDist`.

The fix for this is in `EntityDragger.backOff()`: guard against the zero vector before calling `normalize()`. #1746 includes both the `@WrapOperation` change and the `backOff()` NaN fix.

Separately, the `@Redirect` approach here is exclusive. Only one mod can `@Redirect` the same call site. If another mod also needs to modify `setDeltaMovement(DDD)V` in `Entity.move()`, it'll hard-conflict at launch. `@WrapOperation` composes, so multiple WrapOperations chain without conflict.

hehdawn:
PR superseded by [1770](https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1770)

PR #1766, opened 2026-04-06
Downsize to 2.4.11

blockninja124:
Reverts the mod version from 2.5.0 to 2.4.11, as well as updating the changelog accordingly. This should allow us to do a bug fix update soon:tm: despite air pockets not being done yet.


blockninja124:
Air pockets are entirely not merged to main yet btw, which is what allows this temp release to be done.

Issue #1763, opened 2026-04-03
Crash during world loading


IT-Quant:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods

- [x] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

1.20.x

### Mod Loader

Forge

### Issue description

Trying to launch a world with VS 2.4.6-2.4.10 crashes the game. Installed only VS and Kotlin for forge 4.11. Forge version 47.4.10, mc version 1.20.1

Error: java.lang.ClassCastException: class net.minecraft.world.phys.Vec3 cannot be cast to class net.minecraft.core.BlockPos (net.minecraft.world.phys.Vec3 and net.minecraft.core.BlockPos are in module minecraft@1.20.1 of loader 'TRANSFORMER' @62b57479)

Checked for similar issues, found one, solution didn't work 

Crash report:

[crash-2026-04-03_12.28.36-client.txt](https://github.com/user-attachments/files/26455717/crash-2026-04-03_12.28.36-client.txt)

### Issue reproduction

Load any world with VS installed

### Logs

[latest.log](https://github.com/user-attachments/files/26455775/latest.log)

[debug.log](https://github.com/user-attachments/files/26455780/debug.log)

Daniel99j2:
Bump. Same issue!

blockninja124:
You need to use any launcher other than vanilla.

Curseforge app, modrinth app, prism launcher, multi mc, poly mc, etc all work. Only vanilla is broken.

We haven't been able to figure out why. If you experience this and are already using a custom launcher, then please share your modpack so we can try to narrow down the cause.

Daniel99j2:
Thanks- I’m using Pandora, but given its unintrusiveness, it probably doesn’t fix it.

Daniel99j2:
[options.txt](https://github.com/user-attachments/files/26616959/options.txt)
I found something - using the default options.txt: it loads. Copying an options.txt from my normal 1.21.11 fabric instance crashes!
That would also explain the vanilla launcher not working - by default all instances share the same run directory

blockninja124:
thats very interesting. The file you included, is it the 1.21.11 options.txt? And can this be reproduced using only a default 1.21.11 options.txt (aka absolutely no changes).

Daniel99j2:
Found the issue - prioritizeChunkUpdates:2 causes the crash! (1 or 0 work)

[...]

PR #1743, opened 2026-03-20
fix: Resolve ClassCastException (Vec3 to BlockPos) with vanilla rendering

Pycnocline:
## What this PR does:
- [x] Bug fix 
- [ ] Feature
- [ ] Code refactor / improvement
- [ ] API addition / improvement
- [ ] Compat with another mod

**Explain what this PR is doing:**
This PR fixes a crash that occurs when entering the game under specific conditions while using vanilla rendering. 

The game was throwing the following crash:
`java.lang.ClassCastException: class net.minecraft.world.phys.Vec3 cannot be cast to class net.minecraft.core.BlockPos`

I fixed this by converting the Vec3 world coordinates to BlockPos using BlockPos.containing() before calling the distSqr operation in the distToShips mixin.

---

## Modloaders this PR affects:
- [x] Forge
- [ ] Fabric

## Where this PR was tested:
- [ ] In-dev singleplayer
- [ ] In-dev dedicated server
- [x] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [x] No

---

## Additional Notes
Anything else reviewers might need to know:
Hi! This is my first time contributing to this project. I found this ClassCastException bug and tried my best to fix it here. I'm not 100% sure if this is the most optimal approach for the mod's architecture, so please let me know if there's anything I should change or improve! 

---

## Declaration
By submitting this PR, I affirm:  
- Code/data compiles and loads  
- Tests _(if any)_ pass  
- This PR will not brick worlds

PR #1780, opened 2026-04-14
Update MixinLevelRendererVanilla.java

Daniel99j2:
## What this PR does:
- [x] Bug fix 
- [ ] Feature
- [ ] Compat with another mod

This PR comments out a line which caused the issues described in #1763. The line had the error 'Cannot resolve any target instructions in target class'.

Vanilla code
```
if (this.minecraft.options.prioritizeChunkUpdates().get() == PrioritizeChunkUpdates.NEARBY) { //type 2
    BlockPos blockPos2 = renderChunk.getOrigin().offset(8, 8, 8);
    bl = blockPos2.distSqr(blockPos) < (double)768.0F || renderChunk.isDirtyFromPlayer(); //this call caused the crash
} else if (this.minecraft.options.prioritizeChunkUpdates().get() == PrioritizeChunkUpdates.PLAYER_AFFECTED) {
    bl = renderChunk.isDirtyFromPlayer();
}
```
---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework (I tested this and it failed, however my change doesn't affect it!)

## Breaking Changes
- [ ] Yes (describe)
- [x] No

---

## Additional Notes
The line didn't appear to affect ship rendering, but if a proper VS dev could verify this it would be appreciated.
---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/vanilla_renderer/MixinLevelRendererVanilla.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/mod_compat/vanilla_renderer/MixinLevelRendererVanilla.java
@@ -231,14 +231,14 @@ public abstract class MixinLevelRendererVanilla implements LevelRendererDuck, Le
         }
     }
 
-    // If ships stop rendering try uncommenting this line.
-//    @WrapOperation(
-//        method = "*",
-//        at = @At(value = "INVOKE", target = "Lnet/minecraft/core/BlockPos;distSqr(Lnet/minecraft/core/Vec3i;)D")
-//    )
-//    private double distToShips(BlockPos from, Vec3i to, Operation<Double> distSqr){
-//        return VSGameUtilsKt.squaredDistanceBetweenInclShips(level, from.getCenter(), Vec3.atCenterOf(to), distSqr);
-//    }
+    @WrapOperation(
+        method = "*",
+        at = @At(value = "INVOKE", target = "Lnet/minecraft/core/BlockPos;distSqr(Lnet/minecraft/core/Vec3i;)D"),
+        require = 0
+    )
+    private double distToShips(BlockPos from, Vec3i to, Operation<Double> distSqr){
+        return VSGameUtilsKt.squaredDistanceBetweenInclShips(level, from.getCenter(), Vec3.atCenterOf(to), distSqr);
+    }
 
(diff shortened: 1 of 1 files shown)


Daniel99j2:
Because of the error 'Cannot resolve any target instructions in target class' I suspect this mixin has not actually worked, so it should be safe to merge.

blockninja124:
I don't know if the mixin is required, but it looks like it was added somewhat recently so it might be. There's a safer (and slightly more janky) way to fix the mixin, which is making it `require=0` so that if it fails (if chunk priority is 2) it will silently fail instead of crashing the game.

chj990707:
I think 'Cannot resolve any target instructions' is an error confined to IDE, and not the cause of the crash (besides, I think if that causes crashes it should be on the loading screen). Issue #1763 is about crash during world loading and it is caused by the mixin trying to use a Vec3 for Vec3i. Which can be fixed by following change:
```
@WrapOperation(
        method = "*",
        at = @At(value = "INVOKE", target = "Lnet/minecraft/core/BlockPos;distSqr(Lnet/minecraft/core/Vec3i;)D")
    )
    private double distToShips(BlockPos from, Vec3i to, Operation<Double> distSqr){
        return distSqr.call(BlockPos.containing(VSGameUtilsKt.toWorldCoordinates(level, from)), BlockPos.containing(VSGameUtilsKt.toWorldCoordinates(level, new BlockPos(to))));
    }
```
But as far as I know the purpose of this mixin is to sort the chunk being rendered considering ship distance, because the ship chunks are considered far away without the code, and I think this crash only happening on 'Full-blocking' chunk building behaviour is strange. That would mean the sorting is only done on Full blocking behaviour, or the others are omitted to check that.

chj990707:
Found out that the mixin was supposed to target an instruction that is in lambda statement, which means this mixin was doing nothing intended... and also it was added by me(sorry!). It can be safely removed without commenting it out.

blockninja124:
Supersedes https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1743 and fixes https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/1763

blockninja124 (approved):

PR #1800, opened 2026-04-29
Tweaked lava

blockninja124:
**Do not merge this blindly**
This PR changes the values for Lava, which changes the balance of lava boats and submarines.

It changes them from the previous density of 10,000 (water is 1,000) and a drag factor of 1.0 (water is 0.3) to the new values of a mass of 900 (so its harder to float on than water) but with a drag factor of 1.5 (so its a lot thicker than water)

Tomato suggested this, but these values are just a proposal. They may need a different balance/tweaking.


Tribulla:
I don't think you need this many changes to change how lava works :clueless:

PR #1812, opened 2026-04-30
camelCase -> No More

blockninja124:
Add a small tweak to the previous PR https://github.com/ValkyrienSkies/Valkyrien-Skies-2/pull/1811 which automatically converts the camelCase config names into "Camel Case".

image


blockninja124:
yes `Recommend Slugs In Mc Commands` looks a bit silly, but its better than `recommendSlugsInMcCommands` :P

PR #1827, opened 2026-05-02
1.20.1/fix riding view vector

chj990707:
## What this PR does:
- [x] Bug fix 
- [x] Code refactor / improvement
- [x] Compat with another mod

1. Misalignment of looking direction
When VS renders mounted(seated on ship) entities, it will transform the entities with rotation of the ship.
This caused look vector and rendered looking direction of entities to misalign, so previously this was fixed by also rotating the view vector (and camera rotation of players) of said entities. Although this was fine for many cases, some mechanics and other mods, e.g. compass direction and minimap markers, uses raw values stored for yaw and pitch of entities. Following image is one of such occasions, nameTags and other sprites being misaligned.To resolve this, this PR will preserve the original view vector and instead change the rendered rotation of living entities, by calculating the yaw and pitch from the ship transform and entities' view vector. By this way, other codes can safely refer to consistent view vector and rotation angles, while also getting the rendered rotation right.

2. Order of Entity.pick() in MixinGameRenderer
This is mainly done for Shoulder Surfing reloaded compatibility. Previous code has separately executed raycast operation to get untransformed(in shipyard) location of hit result to use on interactions. But when the Entity.pick() was modified by other mods to be using different raycast trace, rendered target of crosshair and actual position used for interactions didn't match.
This PR will remove the separate raycasting, and use the result of Entity.pick() by creating a copy of block hit result that has untransformed location instead. 
---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] Out of dev singleplayer

## Breaking Changes
- [x] Yes (describe)
---
This PR will change the view vector and rendered rotation of entities mounted to a ship. This will fix many occasions of misalignment in player/entity direction. Also, this can possibly affect other codes that rotates the entities on a ship, causing another bug, although none was found during testing.
Compatible with Shoulder surfing reloaded(modified to remove the incompatibility notation).

## Additional Notes
---
Might break other codes that separately dealt with entity rotation.
Might have priority issue with other mixins to the Entity.pick() method.

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/client/MixinCamera.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/client/MixinCamera.java
@@ -2,6 +2,7 @@ package org.valkyrienskies.mod.mixin.client;
 
 import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import net.minecraft.client.Camera;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.player.LocalPlayer;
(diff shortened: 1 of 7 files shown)


allanonmage:
I came here from the Shoulder Surfing Reloaded side of the fix, but I don't see a new version of VS has been released, so I'm stuck trying to have them both in my modpack only to realize that while it's been fixed, the fix hasn't been rolled out yet. 5 months without a compile/update seems like a long time given the previous one was 2 months prior, and a fairly rapid schedule before that (every 1-2 months). I checked Modrinth and Curseforge. Is there an ETA or anything?

blockninja124:
We do not give ETAs, it is simply impossible to predict our development speed. The large gap between updates is because this upcoming update is rather massive, and still a work in progress. Lots and lots of rewrites, optimizations, and misc changes.

If you want the fix sooner, you could potentially compile a custom jar based off the 2.4.11 release tag but with these changes cherry-picked. I have a tutorial on how to do that [here](https://www.youtube.com/watch?v=t6I8yaHnkLE). You would be free to redistribute that custom jar as part of your modpack, as VS is licensed LGPL-3.0. (Although you may have to argue with Curseforge support if you want to put it on there)

PR #1857, opened 2026-05-13
Pistons pushing ships

### LINKED ISSUE #37: pistons do not push ships
an extending piston doesn't push ships, only the head clipping into it once extended fully does, with great force

copperwarrior:
## What this PR does:
- [X] Feature

Allows pistons to push ships in what should be (in my opinion) a normal and sensible way
Force is scaled by mass and volume to keep pistons powerful but not overpowerful
Multiple pistons applying force to the same ship have their force divided amongst the pistons to prevent easy massive force
Config value to adjust piston target speed

Also adds a helper for ship AABB to world OBB

Fixes #37 

---

- [X] Forge
- [X] Fabric

## Where this PR was tested:
- [X] In-dev singleplayer
- [ ] In-dev dedicated server
- [X] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [ ] No

---

## Additional Notes
For any number of reasons, it's not a perfect solution, sometimes blocks still clip into ship blocks during a push. I personally think it's fine, and better than just assuming the collision will take care of it.

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/piston_ship_push/MixinMinecraftServer.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/piston_ship_push/MixinMinecraftServer.java
@@ -1,0 +1,16 @@
+package org.valkyrienskies.mod.mixin.feature.piston_ship_push;
+
+import net.minecraft.server.MinecraftServer;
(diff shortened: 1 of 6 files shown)


Tribulla:
Should be PRd into playtest if isn't already

PR #1858, opened 2026-05-13
Rewrite ship argument completely

blockninja124:
## What this PR does:
- [ ] Bug fix 
- [x] Feature
- [x] Code refactor / improvement
- [ ] Compat with another mod

This PR completely rewrites the parsing/suggestion logic behind the `ShipArgument` (`@v`) to be more similar to Mojangs entity selector. This also means we now have dynamically registered options for the selector, and I've added a bunch.
We now have:
distance, dx, dy, dz, id, limit, mass, omega, size, slug, sort, static, velocity, x, x_omega, x_rotation, x_velocity, y, y_omega, y_rotation, y_velocity, z, z_omega, z_rotation and z_velocity

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [x] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [x] Yes (describe)
- [ ] No

Any addons using the `ShipSelector` or `ShipSelectorParser` to manually parse ship arguments will find that the signatures have completely changed. However, most (all?) addons only use `ShipArgument`, which has not changed.

---

## Additional notes:

There was some code used from [Industria](https://github.com/M-Marvin/MCMOD-Industria/blob/main-1.20/IndustriaCore-1.20/src/main/java/de/m_marvin/industria/core/physics/engine/commands/arguments/contraption/ContraptionArgument.java), which is licensed under ARR. However, M_Marvin has given me permission to use the code + use the code in an LGPL setting. Additionally, the code had to be modified significantly to better fit into the previous VS command system, so it's mostly the general file/code layout that remains from Industria.

image

## Declaration
- Tests _(if any)_ pass  


blockninja124:
Please **squash and merge** when merging this, the git history is a little messy.

blockninja124:
Already merged to playtest

Issue #520, opened 2023-06-29
<title>When on a ship, you are able to go past the world border


CptCore:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods

- [X] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

1.19

### Mod Loader

Forge

### Issue description

While on a boat, player is able to go past the world border, whether on the helm or just riding the boat.

### Issue reproduction

Create a boat in a world with a world border, sail said boat past the world border

Video showing the process from boat assembly to getting past the border: https://youtu.be/j0eS9fQMr7c

blockninja124:
this is still an issue on latest VS 1.20.1

PR #1873, opened 2026-05-19
World border and void fixes

blockninja124:
## What this PR does:
- [x] Bug fix?
- [x] Feature

This PR makes ships bounce back from the world border, fixing https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/520

https://github.com/user-attachments/assets/2c2c470e-b513-4c7e-b32e-a23948231383

It also makes ships stop falling into the void after a certain point. They will stabilize at 3 blocks below the death line, so you would need another ship to fish them out without dying. The offset from the death line is configurable (server config), and you can also add 
`"disableVoidSaving": true` to a datapacks dimension properties to restore the old behaviour (falling into the void forever) for a particular dimension.

https://github.com/user-attachments/assets/5134bb34-9715-4e36-9b8c-536448fe2747

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [ ] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [x] Yes (describe)
- [ ] No

The constructor for `DimensionParametersResolver.Parameters` changed, although I doubt any addons were using it.

---

## Additional Notes

This has added some extra work to our default shipLoadEvent callback, namely getting some level info (like worldborder and void height) and giving it to the attachments. I think the performance cost is probably negligible, but its worth keeping an eye on.

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSGameConfig.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSGameConfig.kt
@@ -268,6 +268,11 @@ object VSGameConfig {
         )
         var defaultSplitGraceTimer = 2
 
+        @ConfigEntry(
+            description = "The offset (in blocks) from the death-height ships will stop falling at. To disable saving ships from falling into the void, you need a dimension property datapack."
+        )
+        var voidShipOffset = -3
+
         @ConfigCategory(title = "Commands")
         val Commands = COMMANDS()
 
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/util/WorldBorderAttachment.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/util/WorldBorderAttachment.kt
@@ -1,0 +1,44 @@
+package org.valkyrienskies.mod.common.util
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import net.minecraft.world.level.border.WorldBorder
+import org.joml.Vector3d
+import org.valkyrienskies.core.api.ships.PhysShip
+import org.valkyrienskies.core.api.ships.ShipPhysicsListener
+import org.valkyrienskies.core.api.util.PhysTickOnly
+import org.valkyrienskies.core.api.world.PhysLevel
+import kotlin.math.abs
+
+class WorldBorderAttachment : ShipPhysicsListener {
+    @Volatile
+    @JsonIgnore
+    var border = WorldBorder()
+
+    @OptIn(PhysTickOnly::class)
+    override fun physTick(
+        physShip: PhysShip, physLevel: PhysLevel
+    ) {
+        val pos = physShip.transform.positionInWorld
+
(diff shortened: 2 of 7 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSGameConfig.kt
```diff
@@ -268,6 +268,11 @@ object VSGameConfig {
         )
         var defaultSplitGraceTimer = 2
 
+        @ConfigEntry(
+            description = "The offset (in blocks) from the death-height ships will stop falling at. To disable saving ships from falling into the void, you need a dimension property datapack."
+        )
+        var voidShipOffset = -3
```
copperwarrior, line 274: Could this be written as a direct Y value? "Offset from death-height" may not even be relevant if the dimension has a different death height or even no death height.

common/src/main/kotlin/org/valkyrienskies/mod/common/util/WorldBorderAttachment.kt
```diff
@@ -0,0 +1,44 @@
...
+
+            if (distance > 0) return
+            if (abs(distance) < border.damageSafeZone) return
+
+            distance = abs(distance) - border.damageSafeZone
+
+            // Prevent MASSIVE forces when really far from border
+            // Why are we reusing border.damagePerBlock? because it's roughly the same purpose
+            // default damagePerBlock is 0.2, making our max distance force happen at 10 blocks out
+            distance = distance.coerceAtMost(border.damagePerBlock * 50)
```
copperwarrior, line 34: If we're really far from the border, we should maybe teleport the ship closer to back in bounds? Also for if a ship is stuck on terrain past the border.

copperwarrior (changes requested):
Looks good for the most part, just a couple comments as per the review

common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSGameConfig.kt
blockninja124: The minecraft death height is hardcoded to minBuildHeight-64 for all dimensions (unless mixin'd)

common/src/main/kotlin/org/valkyrienskies/mod/common/util/WorldBorderAttachment.kt
blockninja124: Hm good point, but I don't actually know how to teleport physShips

common/src/main/kotlin/org/valkyrienskies/mod/common/config/VSGameConfig.kt
copperwarrior: Right. It's the mixin'd part that I'm concerned about. It's also a matter of clarity to the end user, they could either write "stop at this exact Y value" or they have to go look up the death height if they don't remember and then gather what they actually want from that value.

Issue #1904, opened 2026-06-13
Teleported to shipyard after grabbing with gravitron (Version ZC)


GormAuslander:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods.

- [x] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

- [x] 1.20.x
- [ ] 1.21.x
- [ ] Older (Unsupported)

### Mod Loader

- [ ] Forge
- [ ] NeoForge
- [x] Fabric
- [ ] Quilt

### Expected Behavior

Not what happened

### Actual Behavior

Game state was not allowing interaction with ship after assembly (couldn't place blocks but could break them). Grabbed with gravitron and my player coordinates instantly showed -2 million or something like that and the game crashed.

### How to Reproduce

Build a ship with the VS Sails dedication botte and grab the helm. If the 3rd person doesn't properly move to far away, try placing blocks. If they don't place, grab with gravitron

### Logs

[crash-2026-06-13_16.48.57-server.txt](https://github.com/user-attachments/files/28919837/crash-2026-06-13_16.48.57-server.txt)

GormAuslander:
[crash-2026-06-13_17.06.17-server.txt](https://github.com/user-attachments/files/28919984/crash-2026-06-13_17.06.17-server.txt)
[crash-2026-06-13_17.09.37-server.txt](https://github.com/user-attachments/files/28919983/crash-2026-06-13_17.09.37-server.txt)
[crash-2026-06-13_17.04.33-server.txt](https://github.com/user-attachments/files/28919982/crash-2026-06-13_17.04.33-server.txt)

New issue with the same crash report "Exception generating chunk": assembling the ship causes it to immediately disappear, Teleporting it back, it will flash on screen for a moment and then disappear again, and then after a few repeats, the game crashes

GormAuslander:
[crash-2026-06-13_17.16.04-server.txt](https://github.com/user-attachments/files/28920007/crash-2026-06-13_17.16.04-server.txt)
This one crashed right on assembly, before I even tried to teleport it. I also had disabled DH this time to see if that was the cause

Tribulla:
fixed

Issue #1878, opened 2026-05-25
(krunch_konstant) stacking debug hinges causes them to spin and eventually shipyard themselves


cesium-muncher:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods.

- [x] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

- [x] 1.20.x
- [ ] 1.21.x
- [ ] Older (Unsupported)

### Mod Loader

- [x] Forge
- [ ] NeoForge
- [ ] Fabric
- [ ] Quilt

### Expected Behavior

hinges may droop a little but overall nothing happens

if you put a chair and sit in it:
hinges still just droop a little but overall nothing happens

### Actual Behavior

hinges start spinning and eventually dissapear. (i think they get shipyarded)

if you put a chair and sit in it:
hinges spin and when they disappear you go very far away, i don't know where though.
one time i crashed because i was a few million blocks away from the integer limit,
another time i was at the world border but the border texture was like stretched

### How to Reproduce

place a few blocks, place hinge on top sideways, activate hinge, place hinge on iron block, activate hinge, repeat steps a few times until blocks start spinning. i recomend 3-6 hinges before placing a seat, if you want.

### Logs

https://mclo.gs/oSZ432i
this is a log of my crash when i went above 64 bit integer limit in position

[video demonstration](https://cdn.discordapp.com/attachments/1036397712431652945/1508346298733170808/Screen_Recording_2026-05-24_222929.mp4?ex=6a153475&is=6a13e2f5&hm=bfdb10d40a2995a129055e7bc4fe44c2bc8c52f4352bdb1de4ea184156be50af&)

backend is KRUNCH_KONSTANT

blockninja124:
I'm able to recreate the spinning (on latest playtest), although not the crash part. Putting a seat on it seems to slow the rotation to stable levels.

Thettigerking:
happens to me but only if i use thrusters to make them spin

PR #1910, opened 2026-06-18
Fix yeet related crashes

blockninja124:
## What this PR does:
- [x] Bug fix 

This PR stores the position of every entity as `lastSafePosition`, unless their position goes out of `Level.isInSpawnableBounds`. 

When that happens, it means a ship has yeeted beyond coordinates Minecraft can handle at all, and entity dragging on trying to move the entity to the invalid position will cause numerous crashes, usually with worldgen.

(This fixes the crash from https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/1904 and https://github.com/ValkyrienSkies/Valkyrien-Skies-2/issues/1878, although it does not fix the actual yeeting, only the outcome.)

On detecting the invalid coordinates, with this PR, it will set the entities location back to the `lastSafePosition`, as well as the ship they are being dragged by. It also zeros out the ships velocity and sets it static.

It then will yell at the player in the chat and in the logs
image
image

The crash can be reproduced reliably on previous VS versions, as the VS teleport command doesn't check the coordinates you teleport a ship to.
Simply sit on a ship, and run `/vs teleport ship-slug 123647280.79 160855281.49 -152674385.65` to cause the crash.

I will be patching the VS teleport command so it respects the proper `Level.isInSpawnableBounds` limit, however I haven't done that just yet as it's useful for testing this edgecase.

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [x] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/shipyard_entities/MixinEntity.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/shipyard_entities/MixinEntity.java
@@ -5,7 +5,10 @@ import com.llamalad7.mixinextras.injector.wrapmethod.WrapMethod;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import java.util.Set;
+import net.minecraft.ChatFormatting;
+import net.minecraft.core.BlockPos;
 import net.minecraft.nbt.CompoundTag;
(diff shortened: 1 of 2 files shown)


blockninja124:
This fixes the crash with the ArrayIndexOutOfBoundsException from the Aquifer$NoiseBasedAquifer

Issue #131, opened 2022-08-02
Particles from blocks do not render in real space


copperwarrior:
Particles emitted by blocks like the torch, furnace and redstone dust do not show in realspace where they should show.
Additionally, the campfire's smoke particles emit in real space but fall downwards instead of float upwards?
Tested on Fabric with a ship built by Eureka.

Rubydesic:
fixed in ac988d7e1d8c824a6d0efedc738626bc234c5b63

PR #1928, opened 2026-07-20
Make slugs data driven

blockninja124:
Needs core PR [#131](https://github.com/ValkyrienSkies/vs-core/pull/131) to be merged

## What this PR does:
- [x] Feature

This PR ties in to the generateSlug hook in CommonHooksImpl that I added in core PR [#131](https://github.com/ValkyrienSkies/vs-core/pull/131), and makes slugs generation data-driven instead of a file in VS core.

Slugs are now determined by datapack files, located at `data/<namespace>/vs_slugs/<file>.json`. VS includes a datapack file for this which is identical to the previous nounlist. 

The files act like forge tag files (although they aren't a tag), in the form of:
```typescript
{
   replace?: bool
   values?: list<string | dict>
   remove?: list<string>
}
```

where slugs can either be a plain string:
```json
{
   "values": [
       "apple",
       "peach",
       "cherry"
   ]
}
```

or a dictionary object, specifying where in the slug they can be:
```json
{
   "values": [
      {
          "id": "one",
          "positions": [0]
      },
      {
          "id": "two",
          "positions": [1]
      },
      {
          "id": "three",
          "positions": [2]
      },
   ]
}
```

The `remove` array is used to remove slugs added by other packs without completely `replace`-ing their files. This is inspired by the addition Forge makes to tag files: https://forge.gemwire.uk/wiki/Tags

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [x] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [x] No

Slugs will not change from the users point of view, as the default list of slugs is the same.

---

## Declaration
- Tests _(if any)_ pass  


Tribulla:
this has a few things.
1. you import io.netty.util.internal.ThreadLocalRandom at line 3 in SlugDatapackResolver.kt while the old code used java.util.concurrent.ThreadLocalRando. Nerry's is and internal package class that was deprecated in Netty 4.1 in favour of the JDK one. It does work now but it could break in the future.
2. if there is a empty slug list or a slug list that is broken the ship slug will just be "null" which isnt a big issue but should be looked out for.
3. GsonHelper.parse and json["id"].asString are unguarded so a single malformed file in any datapack throws out of prepare() and takes down the whole reload.
4. Out-of-range `positions` silently voids an entry. "positions": [5] with NOUNS_PER_NAME = 3 matches no position, so the noun is dropped everywhere with no diagnostic. Worth a warn at load.

small things worth noting for cleanup.
1. Both as MutableList<SlugEntry> casts are unnecessary.
2. slugList.clear() is dead. slugList is reassigned 6 lines later.
3. Shadowed entry in addedSlugs.entries.forEach { entry -> ...filter { entry -> ... } }
4. removedSlugs.contains(...) inside a filter is O(n·m).
5. val loader get() = DataLoader. The get() is pointless since DataLoader is an object.

blockninja124:
1. I didn't notice I had imported the wrong ThreadLocalRandom
2. Yeah thats intentional. The only chance of it happening would be if a datapack creator was overriding the VS values, and so they would probably notice the `null` slugs in-dev and not release a broken datapack.
3. Ah, I had a try catch but it got lost at some point
4. I could add a warning

---

1. I was mutating slugList before, but now you're right, it doesn't need to be mutable
2. Same as above
3. I'll rename the second `entry` to `slugEntry`
5. Not much that can be done about that I don't think, but I can turn `removedSlugs` into a HashSet to make it slightly faster
6. I just copied the VS MassDataLoader, you're right its not neccesary.

Tribulla:
a answer to the 2nd one:
it could be a issue for users that install multiple packs that alone work fine but together they dont (pack A adds positional entries, pack B removes enough of them to empty a positions).

blockninja124:
And what do you propose I do instead?

zyxkad:
> And what do you propose I do instead?

just make `placeholder` word as a backup word ig, `null` is also fine imo

Tribulla:
> And what do you propose I do instead?

```kotlin
fun generateSlug(): String {
    if (slugList.isEmpty()) {
        return "unnamed-ship"
    }

    return List(NOUNS_PER_NAME) { position ->
        val choices = slugsByPosition[position].ifEmpty { slugList }
        choices[ThreadLocalRandom.current().nextInt(0, choices.size)].id
    }.joinToString("-")
}
```

blockninja124:
Circle CI appears to be failing to grab _some_ of the VS core dependency. I wonder if thats because core circle ci got interrupted by proguard timing out the process.

blockninja124:
I think I’ve addressed everything and this is good to merge? Only issue is that the core it depends on is not properly on the maven. I may have to push a dummy commit to core to force it to re-publish.

blockninja124:
This builds now, I bumped the core dep to a version thats actually on maven. Circle CI is failing because of the weird "gradle daemon disappeared" issue which isn't the fault of this PR

blockninja124:
https://cosmic-mod-generator.github.io/valkyrienskies/vs-slugs/ couldn't help myself

PR #1934, opened 2026-07-29
Add the ability to search JEI by block property values

blockninja124:
## What this PR does:
- [x] Compat with another mod

This PR allows you to search JEI using a new prefix, `~`, to filter blocks by their physical properties (mass, friction, and elasticity).

By default, if you simply type in `~x` where x is a number (e.g. `~500`) it will filter for blocks of that weight.
However if you type a more complex search in the form: `~(m:x,f:y,e:z)` you can filter by mass `m`, friction `f`, and elasticity `e`. You can mix any number of filters, for example: 
- `~(m:500,f:0.5)` will find blocks that are 500kg with a friction value of 0.5
- `~(f:0.2,e:0.1)` will find blocks with a friction value of 0.2 and elasticity of 0.1

These values also support ranges and inequalities in the forms `<`, `<=`, `>=`, `>` and `..`.

Is this PR incredibly overkill? Yes. Is it cool regardless? Also yes.

Also this feature can be turned off in the config, or the prefix changed from `~` to another character.

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [x] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

---

## Additional Notes

This _does_ require a Mixin into JEI, however as Mixins go it's not particularly invasive. It simply hooks into the initialization of `ElementPrefixParser` to add an extra `PrefixInfo`. 

Most of this PRs code is in `NumericAttributeStorage`, which handles the parsing/sorting/searching of items. `NumericAttributeStorage` extends `ISearchStorage`, a JEI class which is not in the official API but is still usable.

This PR does add two new methods to `MassDatapackResolver`, those being:
```kotlin
fun getBlockStateFriction(blockState: BlockState): Double?
fun getBlockStateElasticity(blockState: BlockState): Double?
```

Ideally those functions should probably be in `BlockStateInfoProvider` or some other abstracted location, but for the time being `MassDatapackResolver` seems to be the only place which contains elasticity/friction data. So thats for a later API change to fix.

---

## Declaration
- Tests _(if any)_ pass  


whimsicalwren:
made to work with #1938 by making the mixin use ClientBlockStateInfo to get values instead of MassDatapackResolver, and disabling jei search if server disables block info.
also moved the config values for jei search and prefix to client config (which is also where jei places configuration for prefix stuff)

PR #1938, opened 2026-08-02
Make block data sync to clients and allow servers to disable syncing

whimsicalwren:
## What this PR does:
- [x] Bug fix 
- [x] Feature
- [ ] Compat with another mod

this pr fixes (implements, since it didn't exist before) clients being able to see block mass on servers by using a packet to sync them on player join. It also allows servers to disable this syncing so that players cannot see block mass (as well as disabling the tooltip)

---

- [x] Forge
- [x] Fabric

## Where this PR was tested:
- [x] In-dev singleplayer
- [x] In-dev dedicated server
- [ ] Out of dev singleplayer
- [ ] GameTest framework

## Breaking Changes
- [ ] Yes (describe)
- [x] No

---

## Additional Notes
Although this contains no breaking changes, #1934 should be updated to use ClientBlockStateInfo for getting block values

---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
@@ -21,7 +21,7 @@ object ClientBlockStateInfo {
      * Set by server and used on client side, disables mass tooltip completely if false
      */
     @ApiStatus.Internal
-    var enabled = false
+    var shouldAddMassTooltip = false
 
     /**
      * Used to register clientside block data. For internal use only.
@@ -40,7 +40,7 @@ object ClientBlockStateInfo {
      */
     @ApiStatus.Internal
     fun disable() {
-        enabled = false
+        shouldAddMassTooltip = false
         idToBlockData.clear()
     }
 }
--- a/common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
@@ -1,0 +1,46 @@
+package org.valkyrienskies.mod.client
+
+import net.minecraft.core.registries.BuiltInRegistries
+import net.minecraft.resources.ResourceLocation
+import net.minecraft.world.level.block.state.BlockState
+import org.jetbrains.annotations.ApiStatus
+
+data class ClientBlockInfo(
+    val mass: Double,
+    val friction: Double,
+    val elasticity: Double,
+)
+
+/**
+ * For getting block mass, friction, and elasticity on the client side
+ */
+object ClientBlockStateInfo {
+    private val idToBlockData: MutableMap<ResourceLocation, ClientBlockInfo> = HashMap()
+
+    /**
+     * Set by server and used on client side, disables mass tooltip completely if false
+     */
+    @ApiStatus.Internal
+    var enabled = false
+
+    /**
+     * Used to register clientside block data. For internal use only.
+     */
+    @ApiStatus.Internal
+    fun registerBlockInfo(id: ResourceLocation, clientBlockInfo: ClientBlockInfo) {
+        idToBlockData[id] = clientBlockInfo
+    }
+
+    fun getBlockInfo(blockState: BlockState): ClientBlockInfo? = idToBlockData[BuiltInRegistries.BLOCK.getKey(blockState.block)]
+
+    fun getBlockInfo(id: ResourceLocation): ClientBlockInfo? = idToBlockData[id]
+
+    /**
+     * Clears data and disables mass tooltip
+     */
+    @ApiStatus.Internal
+    fun disable() {
+        enabled = false
+        idToBlockData.clear()
+    }
+}
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
@@ -35,7 +35,6 @@ import org.valkyrienskies.mod.api_impl.events.RegisterBlockStateEventImpl
 import org.valkyrienskies.mod.client.ClientBlockInfo
 import org.valkyrienskies.mod.common.BlockStateInfoProvider
 import org.valkyrienskies.mod.common.ValkyrienSkiesMod
-import org.valkyrienskies.mod.common.config.MassDatapackResolver.generateShapeFromVoxel
 import org.valkyrienskies.mod.common.hooks.VSGameEvents
 import org.valkyrienskies.mod.common.networking.PacketSyncBlockStateInfo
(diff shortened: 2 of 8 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
```diff
@@ -0,0 +1,46 @@
...
+ * For getting block mass, friction, and elasticity on the client side
+ */
+object ClientBlockStateInfo {
+    private val idToBlockData: MutableMap<ResourceLocation, ClientBlockInfo> = HashMap()
+
+    /**
+     * Set by server and used on client side, disables mass tooltip completely if false
+     */
+    @ApiStatus.Internal
+    var enabled = false
```
blockninja124: Shouldn't it be enabled by default? Also a cheating client could override this anyway

common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
```diff
@@ -0,0 +1,46 @@
+package org.valkyrienskies.mod.client
+
+import net.minecraft.core.registries.BuiltInRegistries
+import net.minecraft.resources.ResourceLocation
+import net.minecraft.world.level.block.state.BlockState
+import org.jetbrains.annotations.ApiStatus
+
+data class ClientBlockInfo(
```
blockninja124, line 8: Why not make VSBlockStateInfo public and use that? Is this easier for networking or something?

common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
```diff
@@ -449,6 +453,36 @@ object MassDatapackResolver : BlockStateInfoProvider {
...
     }
 
+    /**
+     * For internal use only, used to sync blockstate info from server to client and should not be called otherwise.
+     */
+    @ApiStatus.Internal
+    fun syncBlockStates(player: MinecraftPlayer) {
+        logger.info("Syncing ${mcBlockStateToVs.size} blockstates to ${player.uuid}")
+        with(vsCore.simplePacketNetworking) {
+            val blockstate2VS: Map<String, ClientBlockInfo> =
```
blockninja124: wouldn't resourceKey2VS be more accurate since it's a map of String not BlockState

common/src/main/kotlin/org/valkyrienskies/mod/common/config/MassDatapackResolver.kt
```diff
@@ -449,6 +453,36 @@ object MassDatapackResolver : BlockStateInfoProvider {
...
+                    )
+                    id.toString() to clientInfo
+                }
+                    .toMap()
+
+            PacketSyncBlockStateInfo(blockstate2VS).sendToClient(player)
+        }
+    }
+
+    fun clearBlockStates(player: MinecraftPlayer) {
```
blockninja124, line 478: Shouldn't this also be `@ApiStatus.Internal`

common/src/main/kotlin/org/valkyrienskies/mod/common/networking/VSGamePackets.kt
```diff
@@ -70,6 +72,17 @@ object VSGamePackets {
             }
         }
 
+        PacketSyncBlockStateInfo::class.registerClientHandler { syncBsi ->
+            if (syncBsi.blockState2VS.isEmpty()) {
+                ClientBlockStateInfo.disable()
+            } else {
+                syncBsi.blockState2VS.forEach { (id, state) ->
+                    ClientBlockStateInfo.enabled = true
```
blockninja124: You're enabling the block state info for every blockstate sent in a loop. It should be enabled outside the loop, only once.

fabric/src/main/kotlin/org/valkyrienskies/mod/fabric/common/ValkyrienSkiesModFabric.kt
```diff
@@ -270,6 +271,24 @@ class ValkyrienSkiesModFabric : ModInitializer {
...
+                } else {
+                    MassDatapackResolver.clearBlockStates(player)
+                }
+            }
+        }
+
+        ServerPlayConnectionEvents.DISCONNECT.register { handler, server ->
+            if (handler.player is ServerPlayer) {
+                val player: MinecraftPlayer = handler.player.playerWrapper
+                MassDatapackResolver.clearBlockStates(player)
```
blockninja124: Won't they just get cleared when they next join a singleplayer world or server?

forge/src/main/kotlin/org/valkyrienskies/mod/forge/common/ValkyrienSkiesModForge.kt
```diff
@@ -383,6 +386,24 @@ class ValkyrienSkiesModForge {
...
+            } else {
+                MassDatapackResolver.clearBlockStates(player)
+            }
+        }
+    }
+
+    private fun playerLeave(event: PlayerEvent.PlayerLoggedOutEvent) {
+        if (event.entity is ServerPlayer) {
+            val player: MinecraftPlayer = event.entity.playerWrapper
+            MassDatapackResolver.clearBlockStates(player)
```
blockninja124: Won't they just get cleared when they next join a singleplayer world or server?

blockninja124 (changes requested):
Mostly LGTM, I just have a few questions

common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
whimsicalwren: it doesn't really matter whether it's enabled or not as it gets set on leave/join anyways
additionally, this only controls whether mass tooltip is added, so that players see no value instead of just always seeing the default values just to prevent confusion

blockninja124: I see

common/src/main/kotlin/org/valkyrienskies/mod/client/ClientBlockStateInfo.kt
whimsicalwren: i don't need all the values of this, though i could make it public and use that assuming VsiBlockType will play nice with it
that said, the type field in here is actually never used and is basically always null from what i can tell, which means that field could probably be completely removed without issues

whimsicalwren: actually gonna leave this as is, ClientBlockInfo is a bit easier for networking since i don't need all the values of VSBlockStateInfo and i don't want to (althought unlikely) cause mild breaking changes by removing type from VSBlockStateInfo

blockninja124 (approved):

PR #1954, opened 2026-08-11
Don't crash when another mod @Overwrites a vanilla spawn rule

m3t4f1v3:
## Problem

TerraFirmaGreg's Core mod `@Overwrite`s two of the vanilla spawn-rule helpers we hook, which takes the game down at class-load time:

```
InvalidInjectionException: @At("INVOKE") on Husk::vs$shipAwareCanSeeSky with priority 1000
cannot inject into Husk::m_218996_(...) merged by su.terrafirmagreg.core.mixins.common.minecraft.entities.HuskMixin with priority 1000

InvalidInjectionException: @At("INVOKE") on Slime::vs$shipProjectedY with priority 1000
cannot inject into Slime::m_219112_(...) merged by su.terrafirmagreg.core.mixins.common.minecraft.entities.SlimeMixin with priority 1000
```

Two separate things go wrong:

1. `Injector#findTargetNodes` refuses to inject into a method merged by another mixin unless the injecting mixin's priority is *strictly* greater (`InjectionPoint#checkPriority` is `targetPriority < mixinPriority`). At our default 1000 vs their default 1000, that's a hard failure, and it is thrown during prepare — `require` does not soften it.
2. Even past that check, the instruction we wrap is gone. [TFG's overwrites](https://github.com/TerraFirmaGreg-Team/Core-Modern/blob/main/src/main/java/su/terrafirmagreg/core/mixins/common/minecraft/entities/HuskMixin.java) replace the bodies wholesale — their husk rule is just `checkMonsterSpawnRules(...)` (no `canSeeSky`) and their slime rule is just `checkMobSpawnRules(...)` (no `getY`, no `getMaxLocalRawBrightness`, no `new ChunkPos`).

## Fix

Every mixin in `feature.mob_spawning.per_mob` now declares `priority = 1500`, and every injector in it declares `require = 0`:

- the priority lets us apply *after* a default-priority overwrite and wrap it if the vanilla call is still there;
- `require = 0` lets us quietly step aside when it isn't, instead of trading one crash for `Injection validation failed`.

The default `expect = 1` is untouched, so a wrap that finds nothing still logs a warning — a genuine mapping or version regression stays visible.

Stepping aside is the right behaviour rather than a compromise: a mod that overwrote the vanilla spawn rule has deliberately replaced the very check we were making ship-aware, so mobs just spawn by that mod's rules.

Applied to the whole package rather than only `Husk`/`Slime` — all ten mixins there wrap vanilla static spawn-rule helpers, which are a popular `@Overwrite` target, so they share the failure mode. The rationale is written up once in a new `package-info.java`.

## Testing

`:common:compileJava` passes. Change is annotation-only; no behaviour change when no other mod touches these methods.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/src/main/java/org/valkyrienskies/mod/mixin/feature/mob_spawning/per_mob/MixinBatShipSpawn.java
+++ b/common/src/main/java/org/valkyrienskies/mod/mixin/feature/mob_spawning/per_mob/MixinBatShipSpawn.java
@@ -13,7 +13,7 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.valkyrienskies.mod.common.VSGameUtilsKt;
 
-@Mixin(Bat.class)
+@Mixin(value = Bat.class, priority = 1500)
 public abstract class MixinBatShipSpawn {
 
     @WrapOperation(
@@ -22,7 +22,8 @@ public abstract class MixinBatShipSpawn {
             value = "INVOKE",
             target = "Lnet/minecraft/core/BlockPos;getY()I",
             ordinal = 0
-        )
+        ),
+        require = 0
     )
     private static int vs$shipBatY(
         final BlockPos pos, final Operation<Integer> original,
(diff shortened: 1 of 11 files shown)


m3t4f1v3:
fixed by tfg

Issue #1943, opened 2026-08-11
Bug with updating blocks when assembling (Forge only) Playtest ZF


Thettigerking:
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods.

- [x] I have tested this issue and it occurs when no other mods are installed

### Minecraft Version

- [x] 1.20.x
- [ ] 1.21.x
- [ ] Older (Unsupported)

### Mod Loader

- [x] Forge
- [ ] NeoForge
- [ ] Fabric
- [ ] Quilt

### Expected Behavior

When assembling stuff blocks around it should react, so assembling fences should remove the connections between neighboring fences

### Actual Behavior

When assembling ships blocks dont react. View video:
https://cdn.discordapp.com/attachments/1036397712431652945/1536482027921215578/20260810-2110-24.5495679.mp4?ex=6a7c38a5&is=6a7ae725&hm=5a3ac4403d65d15a32a9225e7332d539fce0358df10eae501fba32b04afde0ce&
https://discord.com/channels/244934352092397568/1536482029208608899/1536484297622683719
https://discord.com/channels/244934352092397568/1536482029208608899/1536555572453965874
https://discord.com/channels/244934352092397568/1536482029208608899/1536555376101560400

This redstone updates on fabric for this but the button doesent:
https://discord.com/channels/244934352092397568/1536482029208608899/1536556700843384933

This is the only video that has a similarity between forge and fabric for this bug. All other videos occur on forge. 

### How to Reproduce

Easiest way is to put 2 fences next to eachother so they connect and then assemble 1.

PR #1975, opened 2026-09-03
bug fixes (pr attempt 3)

### LINKED ISSUE #1956: Buttons and pressure plates get stuck down if assembled while pressed
### This issue occurs when only Valkyrien Skies and addons are installed and no other mods.

### Minecraft Version

- [x] 1.20.x
- [ ] 1.21.x
- [ ] Older (Unsupported)

### Mod Loader

- [x] Forge
- [ ] NeoForge
- [ ] Fabric
- [ ] Quilt

### Expected Behavior

When assembling a button or pressure plate while its pressed it should unpress and not be stuck after assembly

### Actual Behavior

View video:
https://discord.com/channels/244934352092397568/1536482029208608899/1536556700843384933

Buttons are stuck down after assembly, same with pressure plates. 

### How to Reproduce

Place a button, right click it, then quickly assemble it. 

### Logs

_No response_

Thettigerking:
## What this PR does:
- [ ] Bug fix 
- [ ] Compat with another mod

1. Fixes the bug with client sided lerping, where when assembling a ship with a offset COM using the area assembler causes it to create the ship a few blocks off visually, and lerp it into place

2. Fixes bugs with create mod belt assembly, where when assembling a belt dupes the items on it. It also fixes the bug where when you assemble a part of the belt causes the belt to split in half becoming extremely buggy and cursed. 

3. Fixes /vs backend engine default setting the backend to krunch_classic. It now sets it to krunch_konstant.  

4. also fixes #1956 now
---

- [ ] Forge
- [ ] Fabric

## Where this PR was tested:
- [ ] In-dev singleplayer

## Breaking Changes
- [ ] No

---

## Additional Notes

This PR was made with AI help (like 10-15% of it). 

due to issue #1943 assembling partial belts on forge it causes the belts to say as ghost blocks, right clicking them removes them. if #1943 is fixed then this will most likely get fixed. It works perfectly on fabric.

This was my 3rd attempt on making a pr since i kept breaking the older prs but i fixed most of the issues now i think
---

## Declaration
- Tests _(if any)_ pass  

### DIFF
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
@@ -63,8 +63,10 @@ import org.valkyrienskies.mod.compat.create.CreateAssemblyCompat
 import org.valkyrienskies.mod.mixin.accessors.server.level.ServerChunkCacheAccessor
 import org.valkyrienskies.mod.util.AIR
 import org.valkyrienskies.mod.util.StructureTemplateFillFromVoxelSet
+import org.valkyrienskies.mod.util.findPendingBlockTicks
 import org.valkyrienskies.mod.util.logger
 import org.valkyrienskies.mod.util.relocateBlock
+import org.valkyrienskies.mod.util.rescheduleBlockTicks
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.locks.LockSupport
 import kotlin.Int.Companion
@@ -303,6 +305,15 @@ object ShipAssembler {
             }
         }
 
+        val pendingTicksByPos = if (removeOriginal) {
+            blocks.mapNotNull { pos ->
+                val ticks = findPendingBlockTicks(level, pos)
+                if (ticks.isNotEmpty()) pos to ticks else null
+            }.toMap()
+        } else {
+            emptyMap()
+        }
+
         // ========== Removing Old Blocks
         if (removeOriginal) {
 
@@ -386,6 +397,13 @@ object ShipAssembler {
             BlockPos(cornerOfShip.x + dx, cornerOfShip.y + dy, cornerOfShip.z + dz)
         }
         initSkyLightForShip(level, moveDestPositions)
+
+        if (pendingTicksByPos.isNotEmpty()) {
+            blocks.forEachIndexed { index, srcPos ->
+                val ticks = pendingTicksByPos[srcPos] ?: return@forEachIndexed
+                rescheduleBlockTicks(level, moveDestPositions[index], ticks)
+            }
+        }
 
         // ========== Resume Chunk Updates
         val timeAtExecution = level.server.tickCount
--- a/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
+++ b/common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
@@ -37,6 +37,7 @@ import org.valkyrienskies.core.api.util.GameTickOnly
 import org.valkyrienskies.core.impl.config.VSCoreConfig
 import org.valkyrienskies.core.internal.ships.VsiServerShip
 import org.valkyrienskies.mod.common.assembly.ShipAssembler.assembleToShipFull
+import org.valkyrienskies.mod.common.BlockStateInfo
 import org.valkyrienskies.mod.common.dimensionId
 import org.valkyrienskies.mod.common.executeIf
 import org.valkyrienskies.mod.common.forEach
@@ -56,6 +57,8 @@ import org.valkyrienskies.mod.common.util.toJOMLD
 import org.valkyrienskies.mod.common.vsCore
 import org.valkyrienskies.mod.common.world.VSTicketType
 import org.valkyrienskies.mod.common.yRange
+import org.valkyrienskies.mod.compat.LoadedMods
+import org.valkyrienskies.mod.compat.create.CreateAssemblyCompat
 import org.valkyrienskies.mod.mixin.accessors.server.level.ServerChunkCacheAccessor
 import org.valkyrienskies.mod.util.AIR
 import org.valkyrienskies.mod.util.StructureTemplateFillFromVoxelSet
@@ -115,6 +118,41 @@ object ShipAssembler {
         return chunkSet
     }
 
+    @JvmStatic
+    private fun computeExpectedCenterOfMass(
+        blocksWithState: List<Pair<BlockPos, BlockState>>,
+        minStructurePos: BlockPos,
+        cornerOfShip: BlockPos,
+        geometricCenterFallback: Vector3d
+    ): Vector3d {
+        var totalMass = 0.0
+        val weightedSum = Vector3d()
+
+        for ((srcPos, state) in blocksWithState) {
+            val mass = BlockStateInfo.get(state)?.first ?: 1.0
+            if (mass <= 0.0) continue
+
+            val dx = srcPos.x - minStructurePos.x
+            val dy = srcPos.y - minStructurePos.y
+            val dz = srcPos.z - minStructurePos.z
+
+            weightedSum.add(
+                (cornerOfShip.x + dx + 0.5) * mass,
+                (cornerOfShip.y + dy + 0.5) * mass,
+                (cornerOfShip.z + dz + 0.5) * mass
+            )
+            totalMass += mass
+        }
+
+        // Fall back to the geometric center of the ship's bounding box (the old, pre-fix behavior) if we
+        // somehow have no usable mass data - this keeps behavior unchanged rather than risking a divide by 0.
+        if (totalMass <= 0.0) {
+            return Vector3d(geometricCenterFallback)
+        }
+
+        return weightedSum.div(totalMass)
+    }
+
     data class AssembleContext(val ship: ServerShip, val fromCenter: Vector3d, val toCenter: Vector3d)
 
     @JvmStatic
@@ -149,7 +187,7 @@ object ShipAssembler {
             toShip.id, level.server.tickCount.toLong()
         )
 
-        val (wasSuccessful, _, toCenter) = moveBlocksFromTo(level, blocks, fromShip, toShip, minB, maxB, toShip.chunkClaim.getCenterBlockCoordinates(level.yRange, Vector3i()))
+        val (wasSuccessful, _, toCenter, expectedCenterOfMass) = moveBlocksFromTo(level, blocks, fromShip, toShip, minB, maxB, toShip.chunkClaim.getCenterBlockCoordinates(level.yRange, Vector3i()))
 
         if (!wasSuccessful) {
             level.shipObjectWorld.deleteShip(toShip)
@@ -158,9 +196,8 @@ object ShipAssembler {
             throw error
         }
 
-        //teleport fn uses COM as center of ship, so it calculates such offset that centerOfShip will be "center" instead
         val posOffset =
-            Vector3d(toShip.inertiaData.centerOfMass)
+            Vector3d(expectedCenterOfMass)
                 .sub(Vector3d(toCenter))
                 .let { fromShip?.shipToWorld?.transformDirection(it) ?: it }
 
@@ -171,7 +208,7 @@ object ShipAssembler {
                 (fromShip?.shipToWorld?.transformPosition(Vector3d(fromCenter)) ?: fromCenter).add(posOffset),
                 fromShip?.transform?.shipToWorldRotation ?: Quaterniond(),
                 Vector3d(scale * oldScale, scale * oldScale, scale * oldScale),
-                toCenter
+                expectedCenterOfMass
             )
         ))
         toShip.isStatic = false
@@ -179,8 +216,13 @@ object ShipAssembler {
         return AssembleContext(toShip, fromCenter, toCenter)
     }
 
-    data class MoveContext(val wasSuccessful: Boolean, val fromCenter: Vector3d, val toCenter: Vector3d)
-    private val failedMove = MoveContext(false, Vector3d(), Vector3d())
+    data class MoveContext(
+        val wasSuccessful: Boolean,
+        val fromCenter: Vector3d,
+        val toCenter: Vector3d,
+        val expectedCenterOfMass: Vector3d = Vector3d()
+    )
+    private val failedMove = MoveContext(false, Vector3d(), Vector3d(), Vector3d())
 
     @JvmStatic
     @OptIn(GameTickOnly::class)
@@ -191,8 +233,17 @@ object ShipAssembler {
         minStructurePos: BlockPos, maxStructurePos: BlockPos,
         toCenter: Vector3i,
         removeOriginal: Boolean = true)
-    : MoveContext {
-        val blocks = blocks.filter { level.getBlockState(it).let{!it.isAir && !it.inAssemblyBlacklist()} }.toSet()
+        : MoveContext {
+
+        if (removeOriginal && LoadedMods.create) {
+            CreateAssemblyCompat.breakSplitBeltChains(level, blocks)
+        }
+
+        val blocksWithState = blocks.mapNotNull { pos ->
+            val state = level.getBlockState(pos)
+            if (!state.isAir && !state.inAssemblyBlacklist()) pos to state else null
+        }
+        val blocks = blocksWithState.map { it.first }.toSet()
         if (blocks.isEmpty()) return failedMove
 
         val fromId = fromShip?.id ?: -1L
@@ -254,6 +305,10 @@ object ShipAssembler {
 
         // ========== Removing Old Blocks
         if (removeOriginal) {
+
+            if (LoadedMods.create) {
+                CreateAssemblyCompat.clearCarriedBeltItemsBeforeRemoval(level, blocks)
+            }
             for (pos in blocks) {
                 level.getBlockEntity(pos)?.let {
                     if (it is Clearable) {
@@ -303,6 +358,10 @@ object ShipAssembler {
 
         val centerOfShip = cornerOfShip.toJOMLD().add(offset)
 
+        val expectedCenterOfMass = computeExpectedCenterOfMass(
+            blocksWithState, minStructurePos, cornerOfShip, centerOfShip
+        )
+
         val structureSettings = StructurePlaceSettings().addProcessor(
             ICopyableProcessor(
                 SingleItemMap(fromId, toShip?.id ?: -1L, -1L) {it},
@@ -348,6 +407,10 @@ object ShipAssembler {
                 }
             }
             VSAssemblyEvents.onPasteAfterBlocksAreLoaded.emit(VSAssemblyEvents.OnPasteAfterBlocksAreLoaded(level, fromShip, toShip, Pair(fromCenter, centerOfShip), eventData))
+
+            if (LoadedMods.create) {
+                CreateAssemblyCompat.fixMovedKineticBlockEntities(level, moveDestPositions)
+            }
             //force update connectivity because this new assemblyslop doesn't update it :(
             if (VSCoreConfig.SERVER.sp.enableConnectivity) {
                 for (pos in chunkPoses) {
@@ -377,7 +440,7 @@ object ShipAssembler {
             }
         }
 
-        return MoveContext(true, fromCenter, centerOfShip)
+        return MoveContext(true, fromCenter, centerOfShip, expectedCenterOfMass)
     }
 
     @JvmStatic
@@ -550,7 +613,13 @@ object ShipAssembler {
 
         // Phase 3: Execute all block moves
         val phase3Start = System.currentTimeMillis()
+        val allKineticFixPositions = ArrayList<BlockPos>()
         for (pending in pendingAssemblies) {
+
+            if (LoadedMods.create) {
+                CreateAssemblyCompat.breakSplitBeltChains(level, pending.blocks)
+            }
+
             // Cache block states during filtering to avoid double getBlockState calls
             val filteredBlocksWithState = mutableListOf<Pair<BlockPos, BlockState>>()
             for (pos in pending.blocks) {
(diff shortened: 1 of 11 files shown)


common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -115,6 +118,41 @@ object ShipAssembler {
...
+        blocksWithState: List<Pair<BlockPos, BlockState>>,
+        minStructurePos: BlockPos,
+        cornerOfShip: BlockPos,
+        geometricCenterFallback: Vector3d
+    ): Vector3d {
+        var totalMass = 0.0
+        val weightedSum = Vector3d()
+
+        for ((srcPos, state) in blocksWithState) {
+            val mass = BlockStateInfo.get(state)?.first ?: 1.0
```
whimsicalwren: we should set mass to the value of VSGameConfig.SERVER.defaultBlockMass if it's null, not 1.0.

common/src/main/kotlin/org/valkyrienskies/mod/util/Logging.kt
```diff
@@ -10,6 +10,7 @@ fun logger(name: String): ClassLogger = ClassLogger(LogManager.getLogger(name))
 @JvmInline
 value class ClassLogger(val logger: Logger) {
     operator fun getValue(thisRef: Any?, property: KProperty<*>): Logger = logger
+   // fun warn(string: String, e: Throwable) {}
```
whimsicalwren: ???

whimsicalwren (changes requested):
seems mildly okay but it's a tiger pr so you never know
assuming all this stuff has actually been tested i suppose it's good

whimsicalwren:
> Fixes /vs backend engine default setting the backend to krunch_classic. It now sets it to krunch_konstant.

you made no changes to this command? why is this. listed as a change this pr makes.

yayEvy:
this pr sucks mods ban this guy

common/src/main/kotlin/org/valkyrienskies/mod/util/Logging.kt
Thettigerking: tribulla told me to remove it but I uh kinda just commented it out in case

Thettigerking:
> > Fixes /vs backend engine default setting the backend to krunch_classic. It now sets it to krunch_konstant.
> 
> you made no changes to this command? why is this. listed as a change this pr makes.

I forgor to re include it when I was fixing my pr lemme fix it

Thettigerking:
wait I accidently clicked request review

common/src/main/kotlin/org/valkyrienskies/mod/util/Logging.kt
Tribulla: I didn't tell you to remove it :sob:
I told your code just discards warnings.

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+
+
+    @JvmStatic
+    fun clearCarriedBeltItemsBeforeRemoval(level: Level, pos: BlockPos) {
+        val blockEntity = level.getBlockEntity(pos)
+        if (blockEntity is BeltBlockEntity && blockEntity.isController()) {
+            blockEntity.getInventory()?.getTransportedItems()?.clear()
```
blockninja124, line 31: Couldn't this cause items to be voided

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
```diff
@@ -0,0 +1,79 @@
...
+        val visitedChainStarts = HashSet<BlockPos>()
+        val chainsToBreak = ArrayList<List<BlockPos>>()
+        for (pos in candidateBlocks) {
+            val state = level.getBlockState(pos)
+            if (!AllBlocks.BELT.has(state)) continue
+
+            var start = pos
+            var startState = state
+            while (true) {
+                val prev = BeltBlock.nextSegmentPosition(startState, start, false) ?: break
```
blockninja124, line 57: This seems like it could be expensive, iterating every segment of the belt for every position. You could maybe end up running over the same belt multiple times?

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
```diff
@@ -0,0 +1,79 @@
...
+            val chain = BeltBlock.getBeltChain(level, start)
+            if (chain.size < 2) continue
+
+            if (!candidateBlocks.containsAll(chain)) {
+                chainsToBreak.add(chain)
+            }
+        }
+
+        //break belt
+        for (chain in chainsToBreak) {
```
blockninja124: Why iterate a second time, why not just break the chains right away? aka no need for chainsToBreak

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -12,10 +14,29 @@ import net.minecraft.world.level.block.entity.BlockEntity
...
 
+fun findPendingBlockTicks(level: Level, pos: BlockPos): List<ScheduledTick<Block>> {
+    val serverLevel = level as? ServerLevel ?: return emptyList()
+    @Suppress("UNCHECKED_CAST")
+    val accessor = serverLevel.blockTicks as LevelTicksAccessor<Block>
+    val container = accessor.`vs$getAllContainers`()[ChunkPos.asLong(pos)] ?: return emptyList()
+    return container.getAll().filter { it.pos() == pos }.collect(Collectors.toList())
+}
+
+fun rescheduleBlockTicks(level: Level, to: BlockPos, pendingTicks: List<ScheduledTick<Block>>) {
```
blockninja124: Why is this relocating of block ticks needed? What problem does this solve?

common/src/main/resources/valkyrienskies-common.mixins.json
```diff
@@ -394,7 +394,8 @@
     "world.level.lighting.LayerLightSectionStorageAccessor",
     "world.level.lighting.LayerLightSectionStorageInvoker",
     "world.level.lighting.LightEngineStorageAccessor",
-    "world.level.lighting.MixinThreadedLevelLightEngine"
+    "world.level.lighting.MixinThreadedLevelLightEngine",
+    "accessors.world.ticks.LevelTicksAccessor"
```
blockninja124: Accessors should be at the top of the mixin list

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -115,6 +121,40 @@ object ShipAssembler {
         return chunkSet
     }
 
+    @JvmStatic
+    private fun computeExpectedCenterOfMass(
+        blocksWithState: List<Pair<BlockPos, BlockState>>,
+        minStructurePos: BlockPos,
+        cornerOfShip: BlockPos,
+        geometricCenterFallback: Vector3d
```
blockninja124: You should just return null instead of returning `geometricCenterFallback` and let whatever is calling this function handle the default value

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -149,7 +189,7 @@ object ShipAssembler {
             toShip.id, level.server.tickCount.toLong()
         )
 
-        val (wasSuccessful, _, toCenter) = moveBlocksFromTo(level, blocks, fromShip, toShip, minB, maxB, toShip.chunkClaim.getCenterBlockCoordinates(level.yRange, Vector3i()))
+        val (wasSuccessful, _, toCenter, expectedCenterOfMass) = moveBlocksFromTo(level, blocks, fromShip, toShip, minB, maxB, toShip.chunkClaim.getCenterBlockCoordinates(level.yRange, Vector3i()))
```
blockninja124: This will break API, changing the method signature to return 4 instead of 3. You should create a new `moveBlocksFromTo` method that returns 4 and leave the old one to not break addons.

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -191,8 +235,17 @@ object ShipAssembler {
...
-        val blocks = blocks.filter { level.getBlockState(it).let{!it.isAir && !it.inAssemblyBlacklist()} }.toSet()
+        : MoveContext {
+
+        if (removeOriginal && LoadedMods.create) {
+            CreateAssemblyCompat.breakSplitBeltChains(level, blocks)
+        }
+
+        val blocksWithState = blocks.mapNotNull { pos ->
+            val state = level.getBlockState(pos)
+            if (!state.isAir && !state.inAssemblyBlacklist()) pos to state else null
```
blockninja124, line 248: unrelated to this PR, but we should allow addons to specify that `state.inAssemblyBlacklist` should not be used for a particular moveBlocksFromTo. I actually never intended for the blacklist to be used in VS by default, addons might _want_ to assemble command blocks and whatnot.

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -252,8 +305,21 @@ object ShipAssembler {
             }
         }
 
+        val pendingTicksByPos = if (removeOriginal) {
```
blockninja124: ugly but whatever

common/src/main/kotlin/org/valkyrienskies/mod/common/item/AreaAssemblerItem.kt
```diff
@@ -61,10 +62,18 @@ class AreaAssemblerItem(
                             createNewShipWithBlocks(BlockPos.containing(center.toMinecraft()), denseSet, level)
                         }
                         else {
-                            ShipAssembler.assembleToShip(level, BlockPos.betweenClosed(lowerCorner, upperCorner).map{ it.mutable() }.toSet(), 1.0)
+                            try {
+                                ShipAssembler.assembleToShip(level, BlockPos.betweenClosed(lowerCorner, upperCorner).map{ it.mutable() }.toSet(), 1.0)
+                            } catch (e: Throwable) {
+                                ASSEMBLY_LOGGER.warn("Assembly failed (possibly nothing left to assemble)", e)
+                                ctx.player?.sendSystemMessage(Component.literal("Nothing to assemble!"))
```
blockninja124: Should be a translatable

common/src/main/kotlin/org/valkyrienskies/mod/common/item/ShipAssemblerItem.kt
```diff
@@ -32,10 +33,19 @@ class ShipAssemblerItem(properties: Properties) : VSItem(properties) {
...
+                } catch (e: Throwable) {
+                    ASSEMBLY_LOGGER.warn("Assembly failed (possibly nothing left to assemble)", e)
+                    null
+                }
+                if (shipData != null) {
+                    ctx.player?.sendSystemMessage(
+                        Component.translatable("command.valkyrienskies.shipify.success_one", shipData.slug)
+                    )
+                } else {
+                    ctx.player?.sendSystemMessage(Component.literal("Nothing to assemble!"))
```
blockninja124: Should be a translatable

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
```diff
@@ -0,0 +1,79 @@
...
+import com.simibubi.create.content.kinetics.base.KineticBlockEntity
+import com.simibubi.create.content.kinetics.belt.BeltBlock
+import com.simibubi.create.content.kinetics.belt.BeltBlockEntity
+import net.minecraft.core.BlockPos
+import net.minecraft.world.level.Level
+
+object CreateAssemblyCompat {
+
+    @JvmStatic
+    fun fixMovedKineticBlockEntities(level: Level, positions: Collection<BlockPos>) {
```
blockninja124: updateMovedKineticBlockEntities would be a better name. Its updating them, not "fixing" them.

blockninja124 (changes requested):

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
Thettigerking: items arent voidedcuz betag happens before this runs

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
Thettigerking: #1956

common/src/main/kotlin/org/valkyrienskies/mod/common/item/AreaAssemblerItem.kt
```diff
@@ -61,10 +62,18 @@ class AreaAssemblerItem(
                             createNewShipWithBlocks(BlockPos.containing(center.toMinecraft()), denseSet, level)
                         }
                         else {
-                            ShipAssembler.assembleToShip(level, BlockPos.betweenClosed(lowerCorner, upperCorner).map{ it.mutable() }.toSet(), 1.0)
+                            try {
+                                ShipAssembler.assembleToShip(level, BlockPos.betweenClosed(lowerCorner, upperCorner).map{ it.mutable() }.toSet(), 1.0)
+                            } catch (e: Throwable) {
```
Tribulla: This can hide partially completed assembly failures. This can cause partially assembled ships that will cause problems. Unlikely to arise but quite bad if it does.

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
Tribulla: That is the only way to do it (create also does it this way) but there is no cap so it will just get stuck in a infinite loop

common/src/main/kotlin/org/valkyrienskies/mod/common/item/ShipAssemblerItem.kt
```diff
@@ -32,10 +33,19 @@ class ShipAssemblerItem(properties: Properties) : VSItem(properties) {
...
+                } catch (e: Throwable) {
+                    ASSEMBLY_LOGGER.warn("Assembly failed (possibly nothing left to assemble)", e)
+                    null
+                }
+                if (shipData != null) {
+                    ctx.player?.sendSystemMessage(
+                        Component.translatable("command.valkyrienskies.shipify.success_one", shipData.slug)
+                    )
+                } else {
+                    ctx.player?.sendSystemMessage(Component.literal("argument.valkyrienskies.ship.area_assembler.nothing_to_assemble"))
```
Tribulla: This sends a translation key as it's error message???

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -12,10 +14,29 @@ import net.minecraft.world.level.block.entity.BlockEntity
...
+    if (pendingTicks.isEmpty()) return
+    val serverLevel = level as? ServerLevel ?: return
+    for (tick in pendingTicks) {
+        val remainingDelay = (tick.triggerTick() - serverLevel.gameTime).coerceAtLeast(0L).toInt()
+        serverLevel.scheduleTick(to, tick.type(), remainingDelay, tick.priority())
```
Tribulla: This can change redstone update order and I probably don't need to explain why that is bad

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -12,10 +14,29 @@ import net.minecraft.world.level.block.entity.BlockEntity
...
+import org.valkyrienskies.mod.mixin.accessors.world.ticks.LevelTicksAccessor
+import java.util.stream.Collectors
 
 val AIR = Blocks.AIR.defaultBlockState()
 
+fun findPendingBlockTicks(level: Level, pos: BlockPos): List<ScheduledTick<Block>> {
+    val serverLevel = level as? ServerLevel ?: return emptyList()
+    @Suppress("UNCHECKED_CAST")
+    val accessor = serverLevel.blockTicks as LevelTicksAccessor<Block>
+    val container = accessor.`vs$getAllContainers`()[ChunkPos.asLong(pos)] ?: return emptyList()
```
Tribulla: Buttons can still get stuck when assembly occurs during scheduled-tick processing

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -12,10 +14,29 @@ import net.minecraft.world.level.block.entity.BlockEntity
...
+    return container.getAll().filter { it.pos() == pos }.collect(Collectors.toList())
```
Tribulla: This is gonna cause severe server stalls

common/src/main/kotlin/org/valkyrienskies/mod/compat/create/CreateAssemblyCompat.kt
```diff
@@ -0,0 +1,77 @@
...
+import net.minecraft.world.level.block.Block
+
+object CreateAssemblyCompat {
+
+    @JvmStatic
+    fun updateMovedKineticBlockEntities(level: Level, positions: Collection<BlockPos>) {
+        for (pos in positions) {
+            val blockEntity = level.getBlockEntity(pos)
+            if (blockEntity is KineticBlockEntity) {
+                blockEntity.warnOfMovement()
```
Tribulla, line 18: This is gonna set every create generators SU output to 0 (including creative generators, waterwheels and more)

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -171,17 +190,30 @@ object ShipAssembler {
...
     private val failedMove = MoveContext(false, Vector3d(), Vector3d())
 
+    data class MoveContextWithCenterOfMass(
+        val wasSuccessful: Boolean,
+        val fromCenter: Vector3d,
+        val toCenter: Vector3d,
+        val expectedCenterOfMass: Vector3d = Vector3d()
+    )
+    private val failedMoveWithCenterOfMass = MoveContextWithCenterOfMass(false, Vector3d(), Vector3d(), Vector3d())
+
```
whimsicalwren, line 213: not sure but i feel like this could be done in a better way (not making a whole new data class) without causing breaking changes

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
```diff
@@ -550,7 +635,13 @@ object ShipAssembler {
 
         // Phase 3: Execute all block moves
         val phase3Start = System.currentTimeMillis()
+        val allKineticFixPositions = ArrayList<BlockPos>()
```
whimsicalwren, line 620: rename to allKineticUpdatePositions?

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
whimsicalwren: in general i feel like rescheduling ticks seems like a bad idea

whimsicalwren (commented):
smallish stuff

common/src/main/kotlin/org/valkyrienskies/mod/common/item/ShipAssemblerItem.kt
Thettigerking: oh I forgot translatable oops

common/src/main/kotlin/org/valkyrienskies/mod/common/item/AreaAssemblerItem.kt
Thettigerking: should I just get rid of the catch?

common/src/main/kotlin/org/valkyrienskies/mod/common/assembly/ShipAssembler.kt
Thettigerking: Would it cause breaking changes?

common/src/main/kotlin/org/valkyrienskies/mod/util/RelocationUtil.kt
```diff
@@ -6,16 +6,24 @@ import net.minecraft.world.Clearable
...
 import net.minecraft.world.level.block.entity.BlockEntity
 import net.minecraft.world.level.block.entity.RandomizableContainerBlockEntity
 import net.minecraft.world.level.block.state.BlockState
+import net.minecraft.world.level.block.state.properties.BlockStateProperties
 import net.minecraft.world.level.chunk.LevelChunk
 import org.valkyrienskies.core.api.ships.ServerShip
 
 val AIR = Blocks.AIR.defaultBlockState()
 
+fun BlockState.resetIfPressedButton(): BlockState {
```
whimsicalwren: not only would this break vanilla behavior by unpressing buttons before they should be, didn't the original issue state that this happened with pressure plates as well?
 
additionally, this seems like a very band-aid fix. my guess is that with blocks like buttons and pressure plates that have powered as a blockstate, when we copy the state we lose the be or whatever that should reset it to its unpressed state after a certain delay.
that means that this probably goes for other blocks that work similarly, like create's placard. we should be able to fix this for all blocks like this ideally, not just buttons and pressure plates

just a guess, i'm on my phone rn so i can't really double-check what i'm saying but this may be worth looking into, maybe in a separate pr so that other changes from this one can merge.

whimsicalwren (commented):
you fixed the issue i complained about looks fine to me idk bro

